PR #1, opened 2019-04-11
Create

oxinabox:
Checklist from https://github.com/JuliaCollections/AxisArraysFuture/issues/1#issuecomment-482077891

- [x] Wrapper Type, that wraps an `AbstractArray` giving names to the indexes. The index named can go in the type params
 - [x] Mapping function: `name2dim(::NamedDimsArray, ::Symbol)::Int` that will take an array with named dimentsions and dimention name, and return its dimension index.
 - [x] Overloads for common functions that take a `dims` argument. Like `sum` and `mean`
       - that return `NamedDimArrays`
 - [x] kwarg style indexing https://github.com/JuliaLang/julia/issues/25631
 - [x] Optional `AbstractArray` methods [see docs](https://docs.julialang.org/en/v1/manual/interfaces/index.html#man-interface-array-1)


src/wrapper_array.jl
```diff
@@ -0,0 +1,81 @@
...
+function order_named_inds(A; named_inds...)
+    keys(named_inds) ⊆ dim_names(A) || throw(
+        DimensionMismatch("Expected $(dim_names(A)), got $(keys(named_inds))")
+    )
+
+
+    inds = map(dim_names(A)) do name
+        get(named_inds, name, :)  # default to slicing
+    end
+end
```
oxinabox: Alternative from @iamed2 

Something that does
```
defaults = (a=:, b=:, c=:)  # In the right order
Tuple(merge(defaults, named_inds))
```

So we would need a function that generates that default.

@iamed2  says this would give compile time sorting,
though I am not sure it would with the Error we throw.
Compile time **hates** things that can throw exceptions.

oxinabox:
probably want to implement a bunch of the optional `AbstractArray` stuff

src/wrapper_array.jl
iamed2: If the error means bail out, compile time is fine with it. I believe it's only a problem when you're catching exceptions or when the exception can change the output.

src/name2dim.jl
```diff
@@ -0,0 +1,47 @@
+
+function NamedNotFoundException(name, cands)
+    return ArgumentError(
+        "No dimensioned called $name exists. Dimension names: $(join(cands, ", "))"
+    )
+end
+
+###################################
+
+# We change the name into a `Val` because that will trigger constant propergation
```
iamed2: This doesn't need to be a val for constant propagation, just for dispatch

src/name2dim.jl
```diff
@@ -0,0 +1,47 @@
...
+# and allow this to resolve at compile time when the `name` is a constant
+name2dim(namemap, name) = _name2dim(namemap, Val(name))
```
iamed2: If you have a trivial NamedTuple where names point to their position in the tuple:
```julia
dims = (foo = 1, bar = 2)
```
Then this function will do what `name2dim` does at compile time if possible:
```
name2dim(dims_nt, name) = first(merge(NamedTuple{(name,), Tuple{Int}}((0,)), dims_nt))
```

src/wrapper_array.jl
oxinabox: I will have to recheck tgat

src/name2dim.jl
oxinabox: Correct comment was incorrect notes from earlier experiments in making this compile time

src/wrapper_array.jl
oxinabox: Seems right

```
julia> @code_llvm (()->NamedDims.order_named_inds((:a,)))()

;  @ none:1 within `#41'
define nonnull %jl_value_t addrspace(10)* @"japi1_#41_14567"(%jl_value_t addrspace(10)*, %jl_value_t addrspace(10)**, i32) #0 {
top:
  %3 = alloca %jl_value_t addrspace(10)**, align 8
  store volatile %jl_value_t addrspace(10)** %1, %jl_value_t addrspace(10)*** %3, align 8
  ret %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 5019490584 to %jl_value_t*) to %jl_value_t addrspace(10)*)
}

julia> @code_llvm (()->NamedDims.order_named_inds((:a,), b=2))()

;  @ none:1 within `#43'
; Function Attrs: noreturn
define nonnull %jl_value_t addrspace(10)* @"japi1_#43_14568"(%jl_value_t addrspace(10)*, %jl_value_t addrspace(10)**, i32) #0 {
top:
  %3 = alloca %jl_value_t addrspace(10)**, align 8
  store volatile %jl_value_t addrspace(10)** %1, %jl_value_t addrspace(10)*** %3, align 8
  %4 = call nonnull %jl_value_t addrspace(10)* @"julia_#order_named_inds_14569"({ i64 } addrspace(11)* nocapture readonly addrspacecast ({ i64 }* @0 to { i64 } addrspace(11)*), %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 4595989792 to %jl_value_t*) to %jl_value_t addrspace(10)*))
  call void @llvm.trap()
  unreachable
}
```

oxinabox:
Julia 1.1 is noticable better at constant propergation such that things happen at compile time.

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
+
+"""
+    name2dim(dimnames, [name])
+
+For `dimnames` being a tuple of dimnames (symbols) for dimenensions.
```
nicoleepp:
```suggestion
For `dimnames` being a tuple of dimnames (symbols) for dimensions.
```

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+If called with just the tuple,
+returns a named tuple, with each name maps to a dimension.
```
nicoleepp:
```suggestion
returns a named tuple, with each name mapped to a dimension.
```

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+function name2dim(dimnames::Tuple, names)
+    # This handles things like `(:x, :y)` or `[:x, :y]`
+    # or via the fallbacks `(1,2)`, or `1:5`
+    return map(name->name2dim(dimnames, name), names)
+end
+
+function name2dim(dimnames::Tuple, dim::Union{Integer, Colon})
+    # This is the fallback that allows `NamedDimsArray`'s to be have dimenstions
```
nicoleepp:
```suggestion
    # This is the fallback that allows `NamedDimsArray`'s to have dimensions
```

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+    # referred to by number. This is required to allow functions on `AbstractArray`s
+    # and that use function like `sum(xs; dims=2)` to continue to work without changes
+    # `:` is the default for most methods that take `dims`
+    return dim
+end
+
+
+"""
+    default_inds(dimnames::Tuple)
+This is the defult value for all indexing expressions using the given dimnames.
```
nicoleepp:
```suggestion
This is the default value for all indexing expressions using the given dimnames.
```

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+If called with just the tuple,
+returns a named tuple, with each name maps to a dimension.
+e.g `name2dim((:a, :b)) == (a=1, b=2)`.
+
+If the second `name` argument is given, them the dimension corresponding to that `name`,
+is returned.
+e.g. `name2dim((:a, :b), :b) == 2`
+If that `name` is not found then `0` is returned.
+"""
+function name2dim(dimnames::Tuple)
```
iamed2:
```suggestion
function name2dim(dimnames::NTuple{N, Symbol}) where N
```
And then you could get rid of the second line of the function

iamed2: Also applies to `default_inds`

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+    order_named_inds(dimnames::Tuple; named_inds...)
+
+Returns the values of the `named_inds`, sorted as per the order they appear in `dimnames`,
+with any missing dimnames, having there value set to `:`.
+An error is thrown if any dimnames are given in `named_inds` that do not occur in `dimnames`.
+"""
+function order_named_inds(dimnames::Tuple; named_inds...)
+    # Note: This code is runnable at compile time if input is a constant
+    # If modified, make sure to recheck that it still can run at compile time
+    keys(named_inds) ⊆ dimnames || throw(
```
iamed2: An alternative to this would be checking to see that `length(full_named_inds) == length(dimnames)`

src/NamedDims.jl
```diff
@@ -1,5 +1,11 @@
 module NamedDims
+using Base: @propagate_inbounds
+using Statistics
 
-greet() = print("Hello World!")
+export NamedDimsArray, name2dim, dim_names
```
rofinn: Seems weird to only partially follow the `_` convention.

src/name_core.jl
rofinn: I might go with `name2dim(dn, [name])` to avoid having a function called `dim_names` and a variable called `dimnames`.

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+
+"""
+    determine_remaining_dim(dimnames::Tuple, inds...)
+Given a tuple of dimension names, e.g.
+and a set of index expressesion e.g `1, :, 1:3, [true, false]`,
+determine which are not dropped.
+Dimensions indexed with scalars are dropped
+"""
+@generated function determine_remaining_dim(dimnames::Tuple, inds)
+    # TODO: This still allocates once, and it shouldn't have to
```
iamed2: I'll note the equivalent code with constants also allocates:
```julia
       function gend()
           v1 = getfield((:a, :b, :c), 1)::Symbol
           v2 = getfield((:a, :b, :c), 3)::Symbol
           Core.tuple(v1, v2)::Tuple{Symbol, Symbol}
       end
```

src/name_core.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+"""
+    determine_remaining_dim(dimnames::Tuple, inds...)
+Given a tuple of dimension names, e.g.
+and a set of index expressesion e.g `1, :, 1:3, [true, false]`,
+determine which are not dropped.
+Dimensions indexed with scalars are dropped
+"""
+@generated function determine_remaining_dim(dimnames::Tuple, inds)
```
iamed2:
```suggestion
@generated function determine_remaining_dim(dimnames::Tuple, inds::Tuple)
```

src/name_core.jl
rofinn:
```suggestion
    # This is the fallback that allows `NamedDimsArray`'s to be have dimensions
```

src/wrapper_array.jl
```diff
@@ -0,0 +1,115 @@
+# `L` is for labels, it should be a `Tuple` of `Symbol`s
```
iamed2:
```suggestion
# `L` is for labels, it should be an `NTuple{N, Symbol}`
```
To clarify that it should be N items long

src/wrapper_array.jl
```diff
@@ -0,0 +1,115 @@
...
+## optional
+Base.IndexStyle(::Type{A}) where A<:NamedDimsArray = Base.IndexStyle(parent_type(A))
+
+Base.length(a::NamedDimsArray) = length(parent(a))
+
+Base.axes(a::NamedDimsArray) = axes(parent(a))
+Base.axes(a::NamedDimsArray, dim) = axes(parent(a), name2dim(a, dim))
+
+
+function Base.similar(a::NamedDimsArray{L}, args...) where L
```
iamed2: This will only work when the returned array from `similar` has the same number of dimensions. Maybe the type signature can reflect this?

src/wrapper_array.jl
```diff
@@ -0,0 +1,115 @@
+# `L` is for labels, it should be a `Tuple` of `Symbol`s
+struct NamedDimsArray{L, T, N, A<:AbstractArray{T,N}} <: AbstractArray{T,N}
```
rofinn: Docstring for the type or constructor?

src/wrapper_array.jl
```diff
@@ -0,0 +1,115 @@
...
+parent_type(::Type{<:NamedDimsArray{L,T,N,A}}) where {L,T,N,A} = A
+Base.parent(x::NamedDimsArray) = x.data
+
+
+"""
+    dim_names(A)
+
+Returns a tuple of containing the names of all the dimensions of the array `A`.
+"""
+dim_names(::Type{<:NamedDimsArray{L}}) where L = L
```
rofinn: I'd also be fine with just calling this `names` and not exporting it...?

nickrobinson251: alternatives: `keys` (like NamedTuples, etc) or `Base.names` 
not sure i prefer either of them to just `names`

src/NamedDims.jl
nickrobinson251: i prefer `a2b` naming only for `Dict`s but I don't have a good alternative and it's only an internal function

src/name_core.jl
oxinabox: Yes

src/wrapper_array.jl
oxinabox: I like `names`

src/NamedDims.jl
oxinabox: so `name_to_dim`?
I am ok this that.

oxinabox: `get_dim`?
`dim`?

src/name_core.jl
oxinabox: Perhaps `dim_labels`?

src/NamedDims.jl
nickrobinson251: i quite to like `dim(names)` and `dim(names, name)` ...but maybe it's scarily short?

src/wrapper_array.jl
oxinabox: I don't think so.
`similar` has *so many* overloads.
And we want to leave them the args alone.

src/NamedDims.jl
oxinabox: We have `axes` it is only slightly shorter.

src/wrapper_array.jl
oxinabox: I like `names`, goes well with `dims`

src/name_core.jl
rofinn: Sure... or maybe `name2dim(names, [labels])`?

oxinabox: now that it is called `dim` this is resolved

src/NamedDims.jl
oxinabox: Can't do `dims`,
it is a kwarg on functions that want  to call this.

e.g.  
`sum(xs::NamedDimArray; dims)=sum(parent(xs), dims(xs, dims))`
does  not work; and
`sum(xs::NamedDimArray; dims)=sum(parent(xs), NamedDimArray.dims(xs, dims))`
is ugly.

oxinabox:
> I feel like you would want to return a NamedDimsArray from operations on a NamedDimsArray

Ah yes, indeed I do.
Did not think `sum` etc through fully.

Looks like I will have to implement more functions for working out dimensions, after operations
but that should make #2 #3 #4 easier at least

oxinabox:
Added dropdims, because I implemented the code for it by mistake (I thought summing over a dimension dropped that dimension, but apparently not (did it used to?))

This is ready for another round of review.

nickrobinson251:
LGTM (bar very minor style things) 

I'd quite like to test on both the 1d and Nd case, but extra tests can always be added later :)

Good work!

src/wrapper_array.jl
iamed2: Based on looking at all the overloads in stdlib/Base, I think it's pretty easy to separate it into two categories: one with dims, and one without. And it would be simple to make this only apply without dims:
```suggestion
function Base.similar(a::NamedDimsArray{L}, args::Type...) where L
```

oxinabox: Why the `...`s?
This can't match anything that is used except the type form.
Which is probably ok,
but in that case `...` isn't needed

iamed2: Some SparseArrays methods take two types. 

It appears the only thing that is ever used that isn't types is dims.

oxinabox: Ok, cool

test/name_core.jl
```diff
@@ -0,0 +1,56 @@
...
+        namemap = dim(manynames)
+        @test keys(namemap) == manynames
+        @test values(namemap) == Tuple(1:length(manynames))
+    end
+    @testset "small case" begin
+        @test dim((:x, :y), :x)==1
+        @test dim((:x, :y), :y)==2
+        @test dim((:x, :y), :z)==0  # not found
+    end
+    @testset "large case that" begin
```
nickrobinson251:
```suggestion
    @testset "large case" begin
```

PR #7, opened 2019-04-16
basic math

oxinabox:
This does array math.
Closes #3 

The + and - will likely need to be redone after #4 
since at least `Base` just defined them to call the Broadcasting equiv.
But they are the simple part of this PR, the  complexity is in `*`

PR #9, opened 2019-04-17
Add Broadcasting

oxinabox:
This is the PR for broadcasting.
I am currently pointing it it #7 
just so the differences can be seen easily.
And so can parallelize reviewing

After #7 is merged I will rebase this and re-open this PR against master

### DIFF
--- a/src/broadcasting.jl
+++ b/src/broadcasting.jl
@@ -1,8 +1,6 @@
 # See: https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting-1
 
-function Base.BroadcastStyle(::Type{<:NamedDimsArray})
-    return Broadcast.ArrayStyle{NamedDimsArray}()
-end
+Base.BroadcastStyle(::Type{<:NamedDimsArray}) = Broadcast.ArrayStyle{NamedDimsArray}()
 
 function Base.similar(
     bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{NamedDimsArray}},
--- a/src/broadcasting.jl
+++ b/src/broadcasting.jl
@@ -4,11 +4,11 @@ Base.BroadcastStyle(::Type{<:NamedDimsArray}) = Broadcast.ArrayStyle{NamedDimsAr
 
 function Base.similar(
     bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{NamedDimsArray}},
-    ::Type{ElType}
-) where ElType
+    ::Type{T}
+) where T
 
     L = broadcasted_names(bc)
-    data = similar(Array{ElType}, axes(bc))
+    data = similar(Array{T}, axes(bc))
     return NamedDimsArray{L}(data)
 end
 
--- a/src/broadcasting.jl
+++ b/src/broadcasting.jl
@@ -9,7 +9,6 @@ function Base.similar(
     ::Type{ElType}
 ) where ElType
 
-    # Scan the inputs for the ArrayAndChar:
     L = broadcasted_names(bc)
     data = similar(Array{ElType}, axes(bc))
     return NamedDimsArray{L}(data)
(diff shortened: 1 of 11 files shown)


src/broadcasting.jl
```diff
@@ -0,0 +1,25 @@
+# See: https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting-1
+
+function Base.BroadcastStyle(::Type{<:NamedDimsArray})
```
rofinn:
```
Base.BroadcastStyle(::Type{<:NamedDimsArray}) = Broadcast.ArrayStyle{NamedDimsArray}()
```
Is under 92 characters.

src/name_core.jl
```diff
@@ -109,6 +109,27 @@ function combine_names(names_a, names_b)
     end
 end
 
+combine_names_longest(name, ::Tuple{}) = name
```
rofinn: This section could use some comments. I'm also not a fan of this interface as I feel like we're basically just reimplementing `merge` for tuples of symbols with some specific `:_` handle, but I guess if it isn't exported then we can iterate on it later.

test/broadcasting.jl
```diff
@@ -0,0 +1,84 @@
+using NamedDims
+using NamedDims: names
+using Test
+
+@testset ".+" begin
+    nda = NamedDimsArray{(:a,)}(ones(3))
+
+    @testset "standard case" begin
+        @test nda .+ nda == 2*ones(3)
```
rofinn: Remember to check spacing around operators.

rofinn: Remember to check spacing around operators and arguments.

test/name_core.jl
```diff
@@ -29,19 +30,26 @@ using Test
...
-    @test combine_names((:a,:_), (:a,:_)) == (:a,:_)
+@testset "combine_names/combine_names_longest" begin
+    @test_throws DimensionMismatch combine_names((:a,), (:a, :b,))
 
-    @test combine_names((:a,:b,:c), (:_,:_,:_)) == (:a,:b,:c)
-    @test combine_names((:a,:_,:c), (:_,:b,:_)) == (:a,:b,:c)
-    @test combine_names((:_,:_,:_), (:_,:_,:_)) == (:_,:_,:_)
+    @test combine_names_longest((:a,), (:a, :b,)) == (:a, :b)
+    @test combine_names_longest((:a,), (:a,:_)) == (:a,:_)
+    @test combine_names_longest((:a,:b), (:a,:_,:c)) == (:a,:b,:c)
```
rofinn: Should probably include a DimensionMismatch test case for when the dimensions for `combine_names_longest((:a, :b), (:b, :a))` just for some self documenting test cases.

src/name_core.jl
oxinabox: We are doing that

This is the purely internal function used for purposes of working out the final name after you broadcast things

src/broadcasting.jl
iamed2:
```
help?> Base.Broadcast.ArrayStyle
  Broadcast.ArrayStyle{MyArrayType}() is a BroadcastStyle indicating that an object behaves as an array for
  broadcasting. It presents a simple way to construct Broadcast.AbstractArrayStyles for specific AbstractArray
  container types. Broadcast styles created this way lose track of dimensionality; if keeping track is important for
  your type, you should create your own custom Broadcast.AbstractArrayStyle.
```
Should we be making a subtype of `AbstractArrayStyle` instead? Keeping dimensionality seems important. Also it would be nice to keep track of the wrapped array type (or at least its `BroadcastStyle`), as we might want to call a different `similar` based on that (even if we don't for now).

oxinabox: This seems to work right now.
So I am for now leaning towards leaving it as it.
If we later find issues with how it currently is, we can change it then.

iamed2: Alright, that's fine.

I think that will probably come up as soon as indexes come into play.

rofinn:
LGTM once the conflicts are resolved. I'm still not a fan of `2ones(2)`, but 🤷‍♂️

oxinabox:
Ok, well I am not actually going to merge this PR against this #7 branch. So I will rebase it off master once #7 is in.

PR #15, opened 2019-04-24
Cleanup

oxinabox:
This redoes the name from combine to unify
and also fixes an inconstient naming of `names_a`
`sourcewalk` is really handy
http://mikeinnes.github.io/MacroTools.jl/stable/sourcewalk/

Also it makes another operation run at compile time.
(which was missed because it was in a seperate parallel, PR)

### DIFF
--- a/test/name_core.jl
+++ b/test/name_core.jl
@@ -1,8 +1,8 @@
 using NamedDims
 using NamedDims:
     names,
-    combine_names,
-    combine_names_longest,
+    unify_names,
+    unify_names_longest,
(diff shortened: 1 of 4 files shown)


test/name_core.jl
```diff
@@ -33,27 +33,27 @@ using Test
 end
 
 @testset "combine_names/combine_names_longest" begin
```
nickrobinson251:
```suggestion
@testset "unify_names/unify_names_longest" begin
```

PR #18, opened 2019-04-24
throw error if a non-matching dimension name is passed in 

### LINKED ISSUE #17: make `dim` throw an  error if invalid dimension name given
Right now, if you attempt to index a `NamedDimsArray`
with a name that is not found, it returns 0.

This in turn results in you getting a variety of fun error messages like:
```
 arraysize: dimension out of range
  Stacktrace:
   [1] size at ./array.jl:154 [inlined]
```

```
ArgumentError: dimension must be ≥ 1, got 0
  Stacktrace:
   [1] reduced_indices(::Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}, ::Int64) at ./reducedim.jl:21
```

Where as we could have a good message like:
```
ArgumentError("Invalid dimension name: $dim, for a Array with dimension names $names")
```

The reason right not that it doesn't is I couldn't workout how to do that without breaking constant propergation.
But now that i  have done a few more things are compile time, I think I can now work it out.

oxinabox:
This closes #17

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -56,7 +56,7 @@ function dim(dimnames::Tuple, name::Symbol)
     dimnum = first(merge(this_namemap, full_namemap))
     dimnum isa Int && return dimnum
     throw(ArgumentError(
-        "Specified name ($(repr(name))), does not match any dimension name ($dimnames)"
+        "Specified name ($(repr(name))) does not match any dimension name ($dimnames)"
     ))
 end
 
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -51,9 +51,13 @@ end
 
 function dim(dimnames::Tuple, name::Symbol)
     # 0-Allocations see: `@btime  (()->dim((:a, :b), :a))()`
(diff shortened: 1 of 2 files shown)


src/name_core.jl
```diff
@@ -51,9 +51,13 @@ end
...
 function dim(dimnames::Tuple, name::Symbol)
     # 0-Allocations see: `@btime  (()->dim((:a, :b), :a))()`
-    this_namemap = NamedTuple{(name,), Tuple{Int}}((0,))  # 0 is default we will overwrite
+    this_namemap = NamedTuple{(name,), Tuple{Symbol}}((:notfound,))  # default we will overwrite
     full_namemap = dim(dimnames)
-    return first(merge(this_namemap, full_namemap))
+    dimnum = first(merge(this_namemap, full_namemap))
+    dimnum isa Int && return dimnum
+    throw(ArgumentError(
+        "Specified name ($(repr(name))), does not match any dimension name ($dimnames)"
```
nickrobinson251:
```suggestion
        "Specified name ($(repr(name))) does not match any dimension name ($dimnames)"
```

PR #19, opened 2019-04-25
Make BroadcastingStyle recursive

oxinabox:
Now since our type is a wrapper type,
it has a wrapper style

 - It needs to overload Materialize rather thasn just `similar` because `Tracker` does not support `setindex` (see also https://github.com/JuliaLang/julia/issues/31844)
 - It needs to add `Tracker` as a dependency, because otherwise we can't set the priority of `NamedDimsStyle` as more important than `TrackedStyle`. Possibly Tracker could be better behaved here somehow ? (@MikeInnes)

Some of the extra stuff is because this is ontop of #18 
Should merge that first


src/broadcasting.jl
```diff
@@ -1,23 +1,94 @@
 # See: https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting-1
 
-Base.BroadcastStyle(::Type{<:NamedDimsArray}) = Broadcast.ArrayStyle{NamedDimsArray}()
+using Base.Broadcast:
+    BroadcastStyle, DefaultArrayStyle, AbstractArrayStyle, Unknown,
+    Broadcasted, broadcasted, materialize
```
oxinabox: I am not sure if we want to move these into NamedDims.jl

src/broadcasting.jl
```diff
@@ -1,23 +1,94 @@
...
     ::Type{T}
-) where T
+) where {S,T}
+    inner_bc = Broadcasted{S}(bc.f, bc.args, bc.axes)
+    data = similar(inner_bc, T)
 
     L = broadcasted_names(bc)
-    data = similar(Array{T}, axes(bc))
     return NamedDimsArray{L}(data)
 end
```
oxinabox: given we define `materialize` below,
I don't know that we _need_ to define `similar` ?

We might want it around since the docs say this is the correct way to implement this kind of thing.

src/broadcasting.jl
```diff
@@ -1,23 +1,94 @@
...
 broadcasted_names(a::AbstractArray) = names(a)
 broadcasted_names(a) = tuple()
+
+
+
+##################################
+# Tracker.jl Compat
+using Tracker
+using Tracker: TrackedStyle, TrackedReal
```
oxinabox:
```suggestion
using Tracker: TrackedStyle
```

src/broadcasting.jl
```diff
@@ -1,23 +1,94 @@
...
     combine_names_longest(a_name, b_name)
 end
 broadcasted_names(a::AbstractArray) = names(a)
 broadcasted_names(a) = tuple()
+
+
+
+##################################
+# Tracker.jl Compat
+using Tracker
```
oxinabox:
```suggestion
```

test/broadcasting_multi.jl
```diff
@@ -0,0 +1,18 @@
...
+@testset "Competing Wrappers" begin
+    nda = NamedDimsArray(ones(4), :foo)
+    ta = TrackedArray(5*ones(4))
+    ndt = NamedDimsArray(TrackedArray(5*ones(4)), :foo)
+
+    arrays = (nda, ta, ndt)
+    @testset "$a .- $b" for (a, b) in Iterators.product(arrays, arrays)
+        a === b && continue
+        @test typeof(nda .- ta) <: NamedDimsArray
+        @test nda .- ta |> parent_typeof <: TrackedArray
```
nickrobinson251: syntax!?<!:{

nickrobinson251:
```suggestion
        @test parent_typeof(nda .- ta) <: TrackedArray
```

src/broadcasting.jl
nickrobinson251: yes, please put all module imports in src/NamedDims.jl :) 
(i understand test imports make sense as file-specifc)

src/broadcasting.jl
```diff
@@ -1,23 +1,93 @@
...
     b_name = broadcasted_names(bs...)
     combine_names_longest(a_name, b_name)
 end
 broadcasted_names(a::AbstractArray) = names(a)
 broadcasted_names(a) = tuple()
+
+
+
+##################################
+# Tracker.jl Compat
```
nickrobinson251: why is Tracker special?

oxinabox: Because we really want to be compatible with Tracker so we can use this with Flux.

nickrobinson251: ah, sorry, i thought `materialze` solved the Tracker case, i missed your second opening
> set the priority of NamedDimsStyle as more important than TrackedStyle

👍

test/broadcasting_multi.jl
nickrobinson251: or just `typeof(parent(nda .- ta)) <: TrackedArray)` and delete the function that's only used here

oxinabox: Yeah, originally this had serval of the different combinations written seperately, so It was useful but now since I am just looping over them, not so much

MikeInnes:
> set the priority of `NamedDimsStyle` as more important than `TrackedStyle`

I could be missing something here, but this seems fishy to me. The typical way this works (for example, with CuArrays) is that tracking always stays on the "outside" – the tracking gets unwrapped first, we go into the adjoint for broadcast, and then we broadcast the unwrapped arrays, which hits their broadcasting style.

Part of the reason I mention this is that with Zygote, you won't have any choice – Zygote's adjoint rules will always take priority and NamedDims will run underneath that. So there would have to be some Tracker-specific reason you need this for it to really make sense.

oxinabox:
> tracking always stays on the "outside" – the tracking gets unwrapped first, we go into the adjoint for broadcast, and then we broadcast the unwrapped arrays, which hits their broadcasting style.

Which is also what NamedDims does.
Because it needs to do that too to support Arrays other than Tracker.
So either way
NamedDims will implement a the following:
https://github.com/invenia/NamedDims.jl/blob/cdea1fe6911a35fa60ce7ccee598c383c0bc1cd1/src/broadcasting.jl#L34-L38

And Tracker will also implement the same.
So they will always have an Ambigious method unless you define the Broadcast style for both of them.
I don't think it hugely matters since both forward everything, and then do there own stuff.

Most things work: for `tr_nda` being a Tracked NamedDimsArray `sum(tr_nda; dims=:foo)` works grear, as does multiplying a `TrackedArray` by a `NamedDimsArray`.
Broadcasting seems to work also.

Som things break, on my quick check just now, getindex, with kwargs, and transpose.
Tracker could reasonably fix those.
```
julia> tra'
Tracked 4×5 LinearAlgebra.Adjoint{Float64,Array{Float64,2}}:
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0

julia> tra[foo=1]
ERROR: MethodError: no method matching getindex(::TrackedArray{…,Array{Float64,2}}; foo=1)
```

One thing that would be more annoying would be the 
[name asserting constructor.](https://github.com/invenia/NamedDims.jl/blob/master/README.md#writing-functions-that-accept-nameddimsarrays-or-abstractarrays)
Which converts `AbstractArrays` into `NamedDimsArray`s,
and asserts that `NamedDimsArray`s have the name you are specifying.
It really handy, and is how one can use it inside a complicated layer in a network,
that is ontop of a chain of dense-layers
to check that your last dimension is what you expect it to be (since the last dimension is always though pre-multiplcation.) And at the same time you get to add in the wildcard first dimensionm so you can refer to them both by their name.
There might be a way to extend some of the longic in https://github.com/invenia/NamedDims.jl/pull/20
to also cover this use-case. But of course it could not be a constructor anymore.
So it would need a new interface.

But anyway, even resolving these problems  will not change much.
We will still need the dependency on Tracker, 
it is just that it would now we will say that Tracker should be on the outside.
and the rest of the broadcast logic would need to remain,
as we need that for StaticArray, DistributedArrays, and any other ArrayType with its own broadcasting rules.


> Part of the reason I mention this is that with Zygote, you won't have any choice – Zygote's adjoint rules will always take priority and NamedDims will run underneath that. 

This is fine. It doesn't matter if adjoint is is performed before dimension name conversion or after.
They are othogonal (as is Tracker's Adjoint).
All that matters is that both are done.

MikeInnes:
Ok, that makes sense. I was only questioning the fact that NamedDims would want to come first, but it makes total sense that regardless of specific order there'd be an ambiguity you need to resolve.

oxinabox:
hmm the current code seems to result in my gradients becoming zero.
Not sure if that is directly from this, or if something else is going wrong

src/broadcasting.jl
```diff
@@ -1,19 +1,56 @@
...
-    ::Type{T}
-) where T
+Base.BroadcastStyle(::NamedDimsStyle{A}, ::NamedDimsStyle{B}) where {A, B} = NamedDimsStyle(A(), B())
+Base.BroadcastStyle(::NamedDimsStyle{A}, b::B) where {A, B} = NamedDimsStyle(A(), b)
+Base.BroadcastStyle(a::A, ::NamedDimsStyle{B}) where {A, B} = NamedDimsStyle(a, B())
+Base.BroadcastStyle(::NamedDimsStyle{A}, b::DefaultArrayStyle) where {A} = NamedDimsStyle(A(), b)
+Base.BroadcastStyle(a::AbstractArrayStyle{M}, ::NamedDimsStyle{B}) where {B,M} = NamedDimsStyle(a, B())
+
+
+function unwrap_broadcasted(bc::Broadcasted{NamedDimsStyle{S}}) where S
```
iamed2: Should have a docstring. Also, does it make sense for there to be Broadcasted within Broadcasted or should this function have a different name than the one it calls?

src/broadcasting.jl
```diff
@@ -1,19 +1,56 @@
...
+    inner_args = unwrap_broadcasted.(bc.args)
```
iamed2: Prefer map when mapping a function over a collection. 
```suggestion
    inner_args = map(unwrap_broadcasted, bc.args)
```

test/broadcasting.jl
```diff
@@ -81,3 +82,17 @@ using Test
         end
     end
 end
+
```
iamed2:
```suggestion
```

src/broadcasting.jl
```diff
@@ -1,19 +1,56 @@
...
+"""
+    NamedDimsStyle{S}
+This is a `BroadcastStyle` for NamedDimsArray's
+It preserves the dimension names.
+`S` should be the `BroadcastStyle` of the wrapped type.
+"""
+struct NamedDimsStyle{S <: BroadcastStyle} <: AbstractArrayStyle{Any} end
+NamedDimsStyle(::S) where {S} = NamedDimsStyle{S}()
+NamedDimsStyle(::S, ::Val{N}) where {S,N} = NamedDimsStyle(S(Val(N)))
+NamedDimsStyle(::Val{N}) where N = NamedDimsStyle{DefaultArrayStyle{N}}()
```
oxinabox: We cannot, it is part of the spec for subtyping `AbstractArrayStyle`

src/broadcasting.jl
oxinabox: Yes, it occurs often.
`Broadcast.flatten` exists to get rid of them but in ourcase it seems easier to do this recursively rather than mess with flatten (which changes between 1.0 and 1.1 I think).

IIRC Identical functions to this exist in RecursiveArrayTools, StaticArrays, DistributedArrays and Tracker

src/compat.jl
```diff
@@ -0,0 +1,51 @@
+##################################
+# Tracker.jl Compat
+using Tracker
+using Tracker: TrackedStyle, TrackedReal
```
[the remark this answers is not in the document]
nickrobinson251: seems okay to me 
(i'm sad about the need for this file at all, ofc)

src/broadcasting.jl
iamed2: Where are the `Val` methods used? Can we kill them?

src/NamedDims.jl
```diff
@@ -12,4 +14,6 @@ include("functions.jl")
 include("functions_dims.jl")
 include("functions_math.jl")
 
+include("compat.jl")
 end # module
+
```
iamed2:
```suggestion
```

src/broadcasting.jl
iamed2: Ahhh I see. Well, we'll soon be able to get rid of that :) https://github.com/JuliaLang/julia/pull/31916

PR #20, opened 2019-04-26
[Speculative] Recursively look for names inside wrapper types

oxinabox:
idk if this is a good idea,

Goal would be for better handling of things like `LowerTriangular{<:NamedDimsArray}`

### DIFF
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -67,9 +67,22 @@ Base.parent(x::NamedDimsArray) = x.data
 
 Returns a tuple of containing the names of all the dimensions of the array `A`.
 """
-names(::Type{<:NamedDimsArray{L}}) where L = L
-names(::Type{<:AbstractArray{T, N}}) where {T,N} = ntuple(_->:_, N)
-names(x::T) where T<:AbstractArray = names(T)
+names(x::NamedDimsArray{L}}) where L = L
+function names(x::T) where T<:AbstractArray
+    inner = parent(x)
+    # things that are not wrapper arrays are there own parents,
+    # so at that point we can give up looking for names
+    # also we have no way to deal with inner arrays of different dimensions, even if they do have names
+    # so give up then too
+    if x === inner || ndims(inner) != ndims(x)
+        return default_names(x)
+    else
+        # recurse to look for names on what was wrapped
+        return names(inner)
+    end
+end
+names(x::Array{T,N}) where {T,N} = default_names(x)  # shortcut of the above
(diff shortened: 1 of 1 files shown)


src/wrapper_array.jl
```diff
@@ -67,9 +67,22 @@ Base.parent(x::NamedDimsArray) = x.data
...
-names(::Type{<:NamedDimsArray{L}}) where L = L
-names(::Type{<:AbstractArray{T, N}}) where {T,N} = ntuple(_->:_, N)
-names(x::T) where T<:AbstractArray = names(T)
+names(x::NamedDimsArray{L}}) where L = L
+function names(x::T) where T<:AbstractArray
+    inner = parent(x)
+    # things that are not wrapper arrays are there own parents,
+    # so at that point we can give up looking for names
+    # also we have no way to deal with inner arrays of different dimensions, even if they do have names
+    # so give up then too
```
nickrobinson251: maybe helpful to add an example (cos it's a bit of a weird case)? e.g.
> parent arrays can have a different number of dimensions,  e.g. the parent of a `Diagonal` is a `Vector`
we have no way to deal with the names in this case so give up then too

src/wrapper_array.jl
```diff
@@ -67,9 +67,22 @@ Base.parent(x::NamedDimsArray) = x.data
...
+    if x === inner || ndims(inner) != ndims(x)
+        return default_names(x)
+    else
+        # recurse to look for names on what was wrapped
+        return names(inner)
+    end
+end
+names(x::Array{T,N}) where {T,N} = default_names(x)  # shortcut of the above
+default_names(x::AbstractArray{T, N}) where {T,N} = ntuple(_->:_, N)
```
nickrobinson251: Are these really the names of an un-named thing? 

Aren't these the names of something which has names, they're just unknown/wildcards

The names of an unnamed thing is `nothing` or maybe `ntuple(i -> nothing, N)`?

nickrobinson251:
I like this as an idea, although I disagree with assigning names (even wildcards) to non-named things. We should only gain wildcards if we cannot know the appropriate new name(s) for a previously named thing

mcabbott:
One vote that this would be nice to have. Surely `x::Diagonal(..., NamedDimsArray{(:k,)})` should have `names(x) == (:k, :k)`. This could be a special case: are there any other wrappers which change `ndims`?

Can this recursive idea be extended to make e.g. `sum(x, dims=:k)` work on anything for which  `:k in names(x)`?

Issue #8, opened 2019-04-17
Troubles with having multiple wildcard dimension names


oxinabox:
This issue will be/was introduced with #7 
I think this is an acceptable issue for now really.
As once you have multiple wildcards then you need to be in the case of a 3+ dimension array to have any nonwild card dimensions left that you would want to refer to by name.
Which is fairly rare.

MWE for no duplicate wildcards.
It mostly works fine,
until you try and access any of the named dimensions.

```
julia> nda = NamedDimsArray{(:_,:_,:c)}(ones(10,20,30));

julia> nda[c=2]
ERROR: duplicate field name in NamedTuple: "_" is not unique
Stacktrace:
 [1] default_inds(::Tuple{Symbol,Symbol,Symbol}) at /Users/oxinabox/.julia/dev/NamedDims/src/name_core.jl:61
 [2] #order_named_inds#5(::Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol},NamedTuple{(:c,),Tuple{Int64}}}, ::Function, ::Tuple{Symbol,Symbol,Symbol}) at /Users/oxinabox/.julia/dev/NamedDims/src/name_core.jl:82
 [3] #order_named_inds at ./none:0 [inlined]
 [4] #order_named_inds#19 at /Users/oxinabox/.julia/dev/NamedDims/src/wrapper_array.jl:134 [inlined]
 [5] #order_named_inds at ./none:0 [inlined]
 [6] #getindex#20 at /Users/oxinabox/.julia/dev/NamedDims/src/wrapper_array.jl:143 [inlined]
 [7] (::getfield(Base, Symbol("#kw##getindex")))(::NamedTuple{(:c,),Tuple{Int64}}, ::typeof(getindex), ::NamedDimsArray{(:_, :_, :c),Float64,3,Array{Float64,3}}) at ./none:0
 [8] top-level scope at none:0

julia> sum(nda; dims=:c)
ERROR: duplicate field name in NamedTuple: "_" is not unique
```

oxinabox:
Originally posted by @mcabbott  in #4
but will be solved by the same solution as to fix the above problem:
 
> I was about to open an issue, but perhaps this isn't expected to work yet... anyway this broadcasting operation gives an error:
> ```
> julia> nda = NamedDimsArray{(:x,:y,:z)}(rand(10,20,30)); # as in readme, but Array not Arrays
> 
> julia> nda .+ rand(20)'
[...]

PR #21, opened 2019-05-07
Allow multiple wildcard names

oxinabox:
I am not happy with this.
It introduces another (particularly evil) `@generated` function,
and two `@pure` functions that are not really pure. But that are probably acceptable because if the generic function methods that they call get overwritten, then a bunch of things will break everywhere in the standard library.

It does however solve #8 
and it continues to run everything to do with names at compile time.

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -107,7 +107,7 @@ Base.@pure function tuple_issubset(lhs::Tuple{Vararg{Symbol,N}}, rhs::Tuple{Vara
         for b in rhs
             found |= a === b
         end
-        !found && return false
+        found || return false
     end
     return true
 end
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -31,39 +31,23 @@ _compile_time_return_hack(::Val{X}) where X = X
 
 
 """
-    dim(dimnames, [name])
+    dim(dimnames, name)
 
 For `dimnames` being a tuple of names (symbols) for the dimensions.
-If called with just the tuple,
-returns a named tuple, with each name mapped to a dimension.
-e.g `dim((:a, :b)) == (a=1, b=2)`.
-
-If the second `name` argument is given, them the dimension corresponding to that `name`,
-is returned.
+and `name` being a one of those names
+This returns the dimension corresponding to that `name`,
 e.g. `dim((:a, :b), :b) == 2`
-If that `name` is not found then an error is thrown.
+If that `name` is not one of the given `dimnames` then an error is thrown.
 """
-function dim(dimnames::Tuple)  # TODO: Remove this ?
-    # 0-Allocations see: `@btime (()->dim((:a, :b)))()`
-    ndims = length(dimnames)
-    return NamedTuple{dimnames, NTuple{ndims, Int}}(1:ndims)
-end
-
-Base.@pure function dim_noerror(dimnames::Tuple{Vararg{Symbol, N}}, name::Symbol) where N
-    # 0-Allocations see: @btime  (()->dim_noerror((:a, :b, :c), :c))()
-    for ii in 1:N
-        getfield(dimnames, ii) === name && return ii
-    end
-    return :notfound
-end
-
 function dim(dimnames::Tuple, name::Symbol)::Int
     # 0-Allocations see: `@btime  (()->dim((:a, :b), :b))()`
     dimnum = dim_noerror(dimnames, name)
-    dimnum isa Int && return dimnum
-    throw(ArgumentError(
-        "Specified name ($(repr(name))) does not match any dimension name ($dimnames)"
-    ))
+    if dimnum === 0
+        throw(ArgumentError(
+            "Specified name ($(repr(name))) does not match any dimension name ($dimnames)"
+        ))
+    end
+    return dimnum
 end
 
 function dim(dimnames::Tuple, names)
@@ -78,6 +62,15 @@ function dim(dimnames::Tuple, d::Union{Integer, Colon})
     # `:` is the default for most methods that take `dims`
     return d
 end
+
+Base.@pure function dim_noerror(dimnames::Tuple{Vararg{Symbol, N}}, name::Symbol) where N
+    # 0-Allocations see: @btime  (()->dim_noerror((:a, :b, :c), :c))()
+    for ii in 1:N
+        getfield(dimnames, ii) === name && return ii
+    end
+    return 0
+end
+
 
 
 """
@@ -100,8 +93,14 @@ function permute_dimnames(dimnames::NTuple{N, Symbol}, perm) where N
     return compile_time_return_hack(new_dimnames)
 end
 
-
+"""
+    tuple_issubset
+A version of `is_subset` sepecifically for `Tuple`s of `Symbol`s, that is `@pure`.
+This helps it get optimised out of existance. It is less of an abuse of `@pure` than
+most of the stuff for making `NamedTuples` work.
+"""
 Base.@pure function tuple_issubset(lhs::Tuple{Vararg{Symbol,N}}, rhs::Tuple{Vararg{Symbol,M}}) where {N,M}
+    N <= M || return false
     for a in lhs
         found = false
         for b in rhs
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -41,19 +41,25 @@ e.g `dim((:a, :b)) == (a=1, b=2)`.
 If the second `name` argument is given, them the dimension corresponding to that `name`,
 is returned.
 e.g. `dim((:a, :b), :b) == 2`
-If that `name` is not found then `0` is returned.
+If that `name` is not found then an error is thrown.
 """
-function dim(dimnames::Tuple)
+function dim(dimnames::Tuple)  # TODO: Remove this ?
     # 0-Allocations see: `@btime (()->dim((:a, :b)))()`
     ndims = length(dimnames)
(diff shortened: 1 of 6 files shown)


src/name_core.jl
```diff
@@ -102,15 +100,16 @@ function permute_dimnames(dimnames::NTuple{N, Symbol}, perm) where N
...
-    default_inds(dimnames::Tuple)
-This is the default value for all indexing expressions using the given dimnames.
-Which is to say: take a full slice on everything
-"""
-function default_inds(dimnames::NTuple{N}) where N
-    # 0-Allocations
-    values = ntuple(_->Colon(), N)
-    return NamedTuple{dimnames, NTuple{N, Colon}}(values)
+
+Base.@pure function tuple_issubset(lhs::Tuple{Vararg{Symbol,N}}, rhs::Tuple{Vararg{Symbol,M}}) where {N,M}
```
nickrobinson251: Why do we need this? What is the behaviour that's not available with `issubset`? (and does the answer to that warrant a comment here?)

```julia
julia> issubset((:a, :b), (:a, :b, :c))
true

julia> issubset((:a, :b, :c), (:a, :b))
false

julia> issubset((:a, :b), (:a, :b))
true
```
versus
```julia
julia> tuple_issubset((:a, :b), (:a, :b, :c))
true

julia> tuple_issubset((:a, :b, :c), (:a, :b))
false

julia> tuple_issubset((:a, :b), (:a, :b))
true
```

src/name_core.jl
```diff
@@ -102,15 +100,16 @@ function permute_dimnames(dimnames::NTuple{N, Symbol}, perm) where N
...
+    for a in lhs
+        found = false
+        for b in rhs
+            found |= a === b
+        end
+        !found && return false
```
nickrobinson251: minor: i find it easier to reason about without the negative
```suggestion
        found || return false
```

src/name_core.jl
```diff
@@ -241,22 +241,20 @@ Given a tuple of dimension names, and either a collection of dimensions,
...
-    n_skip = length(anti_names)
-    ret = ntuple(length(full_names) - n_skip) do ii
-        combined_names[ii + n_skip]  # Skip over the ones we left as the start
-    end
-    return compile_time_return_hack(ret)
+# dropped_dims must be a Tuple type where the values are Int literal for the dimensions being dropped
+@generated function _remaining_dimnames_after_dropping(dimnames::NTuple{N,Symbol}, dropped_dims::Type) where N
+    dropped_dims_vals = dropped_dims.parameters[1].parameters
+    keep_names = [:(getfield(dimnames, $ii)) for ii in 1:N if ii ∉ dropped_dims_vals]
+    return Expr(:call, :compile_time_return_hack, Expr(:tuple, keep_names...))
```
nickrobinson251: I do not understand the evil/magic that is this function

perhaps you can say what advantage there is to writing this as an `@generated` function?

but i'm maybe not placed to review here

test/broadcasting.jl
```diff
@@ -80,4 +80,11 @@ using Test
             @test names(total) == (:foo, :bar)
         end
     end
+
+    @testset "https://github.com/invenia/NamedDims.jl/issues/8#issuecomment-490124369" begin
```
nickrobinson251: can we have a better name for this? 
and then include the link as a comment?

oxinabox: I honestly can;t think of a better one.
It could be called `Regression test for #8b` ?

src/name_core.jl
oxinabox: Wring this as a generated function like this was the only way I could avoid allocations.
The particular avantage is that we basically use the `dropped_dims` type 
to exactly specify, at compile time,
the tuple we are returning,
in terms of which elements of the `dimnames` it contains
and then the `compile_time_return_hack` causes those to constant propegate
so this just all happens completely atc compiletime

src/name_core.jl
oxinabox: A good question.
The answer is: this one is tagged `@pure`
and it is a fairly safe use of `@pure`.
(According to discussion I had in JuliaSlack #internal yesterday)
The purity means that we get to constant propegate

src/name_core.jl
```diff
@@ -41,19 +41,25 @@ e.g `dim((:a, :b)) == (a=1, b=2)`.
...
-    # 0-Allocations see: `@btime  (()->dim((:a, :b), :a))()`
-    this_namemap = NamedTuple{(name,), Tuple{Symbol}}((:notfound,))  # default we will overwrite
-    full_namemap = dim(dimnames)
-    dimnum = first(merge(this_namemap, full_namemap))
+Base.@pure function dim_noerror(dimnames::Tuple{Vararg{Symbol, N}}, name::Symbol) where N
+    # 0-Allocations see: @btime  (()->dim_noerror((:a, :b, :c), :c))()
+    for ii in 1:N
+        getfield(dimnames, ii) === name && return ii
+    end
+    return :notfound
```
iamed2: This should be `0`; we can check in the function below if `dimnum === 0` and I think that'll work.

src/name_core.jl
```diff
@@ -41,19 +41,25 @@ e.g `dim((:a, :b)) == (a=1, b=2)`.
 If the second `name` argument is given, them the dimension corresponding to that `name`,
 is returned.
 e.g. `dim((:a, :b), :b) == 2`
-If that `name` is not found then `0` is returned.
+If that `name` is not found then an error is thrown.
 """
-function dim(dimnames::Tuple)
+function dim(dimnames::Tuple)  # TODO: Remove this ?
```
iamed2:
```suggestion
function dims(dimnames::Tuple)
```
Is this name already taken? I think this is a useful function but it is unlike `dim`.

src/name_core.jl
```diff
@@ -120,18 +119,21 @@ Returns the values of the `named_inds`, sorted as per the order they appear in `
 with any missing dimnames, having there value set to `:`.
 An error is thrown if any dimnames are given in `named_inds` that do not occur in `dimnames`.
 """
-function order_named_inds(dimnames::Tuple; named_inds...)
-    # 0-Allocations
+function order_named_inds(dimnames::Tuple{Vararg{Symbol,N}}; named_inds...) where {N}
+    # 0-Allocations: see `@code_typed (()->order_named_inds((:a, :b, :c), (b=1, c=2)))()`
```
iamed2: I am surprised to see this!

src/name_core.jl
iamed2: I am skeptical that this is the only way. I am going to try some things.

iamed2: Alright I couldn't get better things but we should have an issue to make this better

src/name_core.jl
nickrobinson251: please can you add a comment to the code explaining that?

src/name_core.jl
oxinabox: It is not taken.
We just can't use it easily it things that have the kwarg `dims`.

But I do not think tthe `dims` function is that useful.
Not in its current form
Particularly since it breaks if you have multiple wildcard dimensions.
Which everything else in the package supports now as of this PR.
(except things that take the dimensions as kwargs like `getindex` but that is reasonably not to support since it has no canonical answer).

src/name_core.jl
oxinabox: 😁  NamedTuples are not *that* magic.

src/name_core.jl
iamed2: Functions that have that kwarg would just have to qualify it in that scope. Not a big deal; we wouldn't get conflict warnings or anything.

I'm not really happy with the wildcard stuff. But for now you're the only user and it helps you so better to have it than not.

oxinabox: I didn't like the wildcard stuff and first, but I am convinced it is essential to allow painless interop with other array types and with operations like transpose.

Though we might get rid of transpose at least from common usage if we add automatic dimension permuta tion. And we might even be able to cansually force all partially named NamedDimsArrays to only have their wildcards at the back.
Need to open an issue to discuss this though

oxinabox: > Functions that have that kwarg would just have to qualify it in that scope. Not a big deal; we wouldn't get conflict warnings or anything.

Correct.
But I still don't think the function that returns a NamedTuple is that useful

src/name_core.jl
oxinabox: what is the advantage?
Type-stability for if not running at compile time?

src/name_core.jl
iamed2: > But I still don't think the function that returns a NamedTuple is that useful

Yeah okay. Part of me will die though. RIP NamedTuples for everything

Maybe you'll be able to convince me that we need a MultiNamedTuple

iamed2:
Test failure on 32-bit, needs fixing

iamed2:
https://github.com/FluxML/Tracker.jl/pull/28

src/name_core.jl
oxinabox: I had a lot of fun using NamedTuple Tricks.
But alas this PR removed all of them

iamed2: Stupid features getting in the way of elegance

oxinabox:
@iamed2  shall we merge? we know that tracker will tag a release and fix this on 32Bit at some point

iamed2:
We don't know that there aren't other issues that happen on 32-bit that would only show up when the tracker problem is fixed. Have Tracker.jl tag a patch release, since there's literally no downside.

PR #23, opened 2019-05-13
move keep_dims_ind helper functions to be defined before generated function

oxinabox:
It is a rule of generated functions that they can only call things defined before their own definition.
So depending on the order of function calls without this change world-age `MethodError`s can be thrown.


oxinabox:
when I wrote it I wasn't sure that integers were the types that needed to be kept.
It could be renamed to `is_integer_type`

nickrobinson251:
LGTM - Thanks!

PR #24, opened 2019-05-13
Add LinearAlgebra: lu, lq, svd, qr

### LINKED ISSUE #12: unary LinearAlgebra operations
Can we just delegate them all to `parent`?

oxinabox:
This will close #12 
or at least get closer to it.

Right now I just have `lu!` (which gives `lu` and `det` and `logdet`).
I thought it might be good the share now to get feedback on the design of how to represent the factorization types.
This needs tests.


oxinabox:
The same trick can be done for SVD, but it is a bit gross,
as it will attach the same labels to `U` as to `V`,
which will be wrong for both of them
but then during `getproperty` we can correct them by inserting the wildcards as required.

oxinabox:
The fact that we often store the dimension names via settign the type of `factors` or some arbitary field of the factorization is kinda hacky.
It will cause things to break hilariously if anyone uses `getfield` to access the fields of the Factorization.

I think it is fine, because LinearAlgegra basically accesses the underlying fields still via `getproperty`
and we always either strip the names or return the correct names from getproperty.
Infact we might even be able to do better by adding names to things that don't normally have them in `getproperty` to make more underling math work out.
even more danger-zone there though.

Alternative might be to have a `NamedFactorization` type, or something like that

test/functions_linearalgebra.jl
```diff
@@ -0,0 +1,67 @@
...
+@testset "svd" begin
+    nda = NamedDimsArray{(:foo, :bar)}([1.0 2; 3 4])
+    x = svd(nda)
+    @test names(x.U) == (:foo, :_)
+    @test names(x.S) == (:_,)
+    @test names(x.V) == (:_, :bar)
+    @test names(x.Vt) == (:bar, :_)
+
+    # Identity operation should give back original nam,es
+    @test names(x.U * Diagonal(x.S) * x.V) == (:foo, :bar)
```
nickrobinson251: isn't the input recovered by the operation `U * Diagonal(S) * Vt`? i.e. the names of `V` and `Vt` are flipped in this PR?

test/functions_linearalgebra.jl
```diff
@@ -0,0 +1,67 @@
...
+            @test names(x.Q * x.R) == (:foo, :bar)
+
+            pivot && @testset "pivoted" begin
+                @test x isa QRPivoted
+                @test names(x.p) == (:foo,)
+                @test names(x.P) == (:foo, :foo)
+
+                # Identity operation should give back original names
+                @test names(x.P * nda) == (:foo, :bar)
+                @test names(nda[x.p, :]) == (:foo, :bar)
```
nickrobinson251: I think `p` are column pivots and this only passes becuase we test here on a square matrix i.e. should be `nda * x.P` and `nda[:, x.p]` (these are the things recovered by `Q*R`)

and we should proabably also test
```
names(x.Q * x.R * x.P') == (:foo, :bar)
```

test/functions_linearalgebra.jl
```diff
@@ -0,0 +1,67 @@
...
+    nda = NamedDimsArray{(:foo, :bar)}([1.0 2; 3 4])
+    x = lu(nda)
+    @test names(x.L) == (:foo, :_)
+    @test names(x.U) == (:_, :bar)
+    @test names(x.p) == (:foo,)
+    @test names(x.P) == (:foo, :foo)
+
+    # Idenity opperations should give back original names
+    @test names(x.P * nda) == (:foo, :bar)
+    @test names(x.L * x.U) == (:foo, :bar)
```
nickrobinson251: I find the "Identity operations" comment a little misleading, since really `x.P * nda ≈ x.L * x.U` i.e `nda ≈ x.P * x.L * x.U` (becuase `I = S.P * S.P`)

oxinabox:
I should rebase this.

rofinn:
@oxinabox If you wanna rebase your branch I can work on addressing the comments. I think making a `NamedFactorization` type makes sense to me.

oxinabox:
I will rebase this tomorrow then and you can take it over.

oxinabox:
@rofinn  done. That was a weirdly hard rebase

test/functions_linearalgebra.jl
rofinn: Yes, I believe the names should be on the first dimension for `U` and `V` based on this visualization.

https://en.wikipedia.org/wiki/File:Singular_value_decomposition_visualisation.svg

test/functions_linearalgebra.jl
rofinn: Fun fact, that test fails because `dimnames(X.P)` should equal `(:foo, :_)` in order to avoid a `DimensionMismatch` error.

rofinn: Yeah, looking through the QR wiki page it looks like we want `dimnames(x.P) == (:bar, :bar)`?

rofinn:
Alright, now that my changes have been merged back I'm happy to approve, but maybe @nickrobinson251 wants to give it one more look over?

rofinn:
Should we just merge this as is and then I can update AxisKeys.jl?

oxinabox:
yes?
I thought you had taken over this PR.

rofinn:
Yeah, cause Nick was the original reviewer for this PR.

nickrobinson251:
Oh sorry I wasn't aware this was waiting on me. I'll take a look tomorrow :)

test/functions_linearalgebra.jl
```diff
@@ -0,0 +1,137 @@
+using LinearAlgebra
+using NamedDims
+using NamedDims: dimnames
+using Test
+
+function baseline_tests(fact, identity)
```
nickrobinson251, line 6: do we have any tests for the `iterate` methods defined in this PR?

nickrobinson251 (approved):
LGTM.

One question about tests for `iterate` (but i might just have missed them)

test/functions_linearalgebra.jl
rofinn: Not explicitly, but there's a `destructing via iteration` test on line 22. `...` is calling iterate under-the-hood which is what prompted me to implement the `iterate` API for these types. If we think that the behaviour of `...` might change in the future I could add an explicit `iterate` test?

nickrobinson251: Nah, that's fine. So long as the iterate methods are being hit in tests :)

PR #26, opened 2019-05-14
More matmuls, more easily

### LINKED ISSUE #25: matmul with Diagonal is ambigous
I Imagine am going to have a few of these (already have one for `TrackedArray`).
Might want to make a macro to declare the correct result.
Or at least make it a simple backing function

oxinabox:
closes #25

### DIFF
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -42,14 +42,18 @@ It should be used at the top level of a module.
 macro declare_matmul(MatrixT, VectorT=nothing)
     dim_combos = VectorT === nothing ? ((2,2),) : ((1,2), (2,1), (2,2))
     codes = map(dim_combos) do (NA, NB)
-        TA_named = :(NamedDimsArray{<:Any, <:Any, $NA})
-        TB_named = :(NamedDimsArray{<:Any, <:Any, $NB})
+        TA_named = :(NamedDims.NamedDimsArray{<:Any, <:Any, $NA})
+        TB_named = :(NamedDims.NamedDimsArray{<:Any, <:Any, $NB})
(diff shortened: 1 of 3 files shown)


src/functions_math.jl
```diff
@@ -22,17 +22,38 @@ function matrix_prod_names(a, b)
...
+macro declare_matmul(MatrixT, VectorT=nothing)
+    dim_combos = VectorT === nothing ? ((2,2),) : ((1,2), (2,1), (2,2))
+    codes = map(dim_combos) do (NA, NB)
+        TA_named = :(NamedDimsArray{<:Any, <:Any, $NA})
+        TB_named = :(NamedDimsArray{<:Any, <:Any, $NB})
+        TA_other = (VectorT, MatrixT)[NA]
+        TB_other = (VectorT, MatrixT)[NB]
+
+        quote
+            Base.:*(a::$TA_named, b::$TB_other) = *(a, NamedDimsArray{names(b)}(b))
```
christopher-dG: I doubt that this macro will be used anywhere but in this package, but it would be safer to do `NamedDims.NamedDimsArray` just in case.

oxinabox: It might actually.
Since it is required to give compat between two wrapper matrix types.

oxinabox:
hmm there is a GenericArray in tests for this,
and that test would catch what @christopher-dG  identified.
But also that would be time,
and this is fairly well covered in general by the 3 uses of the macro in the code.
So I am thinking not.

oxinabox:
It would be suitably annoying to add the extra tests because they would have to be done in their own module, which does `Using NamedDims: NamedDims` for it to be worth anything, so I think I will not

PR #27, opened 2019-05-15
[speculative] Use Base.names

### LINKED ISSUE #22: Used `Base.names`
After some reflection, I think we should overload `Base.names`
DataFrames does it, and we have a way better argument than they do.

oxinabox:
Closes #22

However, right now this means doing type-piracy,
to make `Base.names(::AbstractArray) = (:_, :_)`
This can (and should) be removed,
it is only really used in a few places, and we can have a internal `_names` if we really nead it.

However, this approach is basically mutally exclusive with #20 
as that actually requires attacheding `names` to `AbstractArray`s.
Though perhaps it can be done in a type-parametric way that avoids type-piracy?

### DIFF
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -49,10 +49,10 @@ macro declare_matmul(MatrixT, VectorT=nothing)
 
         quote
             function Base.:*(a::$TA_named, b::$TB_other)
-                return *(a, NamedDims.NamedDimsArray{NamedDims.names(b)}(b))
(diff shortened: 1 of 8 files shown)


iamed2:
Type piracy isn't worth it IMO

mcabbott:
One vote for either doing this (without piracy, by `names(x)` not working for things without names) or else changing to `dimnames(x)` which can be exported.

oxinabox:
we now export `dimnames` so this is resolved

PR #28, opened 2019-05-17
comment out docs link

oxinabox:
Currently there are no docs worth linking to and the deploy key is not setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 # NamedDims
 
+<!--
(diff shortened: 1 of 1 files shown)


iamed2:
booooo

iamed2:
By default, autodocs is set up which will pull your public docstrings. That's better than nothing. Just set up the deploy key.

oxinabox:
Orly? I did not know that.
Yes ok, will do

iamed2:
Yeah! It's a pretty nice default thingy :)

PR #38, opened 2019-06-05
Add Strided Array Interface

oxinabox:
Because right now StridedArray is a  Union and thus not extensible
(see https://github.com/JuliaLang/julia/pull/30432)
you don't need to implement `strides` explictly for all new strided array types,
which we can assume (and then be proven wrong with later method errors) 
that our wrapped type is.

This is needed for some BLAS stuff called by Distributions.jl

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -2,16 +2,15 @@
 
 ## Helpers:
 
-# if reducing over `:` then results is a scalar
-function nameddimsarray_result(original_nda, reduced_data, reduction_dims::Colon)
-    return reduced_data
-end
-
 function nameddimsarray_result(original_nda, reduced_data, reduction_dims)
     L = dimnames(original_nda)
     return NamedDimsArray{L}(reduced_data)
 end
 
+# if reducing over `:` then results is a scalar
+function nameddimsarray_result(original_nda, reduced_data, reduction_dims::Colon)
+    return reduced_data
+end
 
 ################################################
 # Overloads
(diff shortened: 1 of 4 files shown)


src/functions.jl
```diff
@@ -2,15 +2,16 @@
...
+
 function nameddimsarray_result(original_nda, reduced_data, reduction_dims)
     L = names(original_nda)
     return NamedDimsArray{L}(reduced_data)
 end
 
-# if reducing over `:` then results is a scalar
-function nameddimsarray_result(original_nda, reduced_data, reduction_dims::Colon)
-    return reduced_data
-end
```
nickrobinson251: why move this?

test/wrapper_array.jl
```diff
@@ -146,3 +146,9 @@ end
         @test names(ndb) == (:w, :x,)
     end
 end
+
+@testset "Strided Array Interface" begin
+    nda = NamedDimsArray{(:a, :b)}(ones(3,5))
+    @test strides(nda) == (1, 3)
+    @test stride(nda, :b) == 3 == stride(nda, 2)
```
nickrobinson251: it'd be good to test against the result for the parent array too
```julia
    x = ones(3, 5)
    nda = NamedDimsArray{(:a, :b)}(x)
    @test strides(nda) == (1, 3) == strides(x)
    @test stride(nda, :b) == 3 == stride(nda, 2) == stride(x, 2)
```

oxinabox:
I wanted this again today.
I should come back to this PR

PR #44, opened 2019-06-21
Correct names for single input cov/cor

### LINKED ISSUE #41: Cov result looks confusing
By taking covariance of a 2D NamedDimsArray (A & B), I would expect the result to be one that is transformed to the other dimension i.e. (B & B).  For example,

```
julia> using Statistics, NamedDims

julia> stock_returns = NamedDimsArray{(:return, :stock)}(rand(-0.1:0.0001:0.1, 30, 3));

julia> typeof(stock_returns)
NamedDimsArray{(:return, :stock),Float64,2,Array{Float64,2}}

julia> cov(stock_returns, dims = :return)
3×3 NamedDimsArray{(:return, :stock),Float64,2,Array{Float64,2}}:
 0.00334097    8.62305e-5    0.000456623
 8.62305e-5    0.00406993   -0.000750522
 0.000456623  -0.000750522   0.00397614 
```

I would expect the result to have type:
```
3×3 NamedDimsArray{(:stock, :stock),Float64,2,Array{Float64,2}}:
```

nickrobinson251:
- closes #41 
- NamedDims still only supports single-arg `cor`/`cov` i.e. for single input `X::AbstractMatrix` (or `x::AbstractVector`, which we get for free)

### DIFF
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -67,3 +67,25 @@ function Base.inv(nda::NamedDimsArray{L, T, 2}) where {L,T}
     names = reverse(L)
     return NamedDimsArray{names}(data)
 end
+
+# Statistics
+for fun in (:cor, :cov)
+    @eval function Statistics.$fun(a::NamedDimsArray{L, T, 2}; dims=1, kwargs...) where {L, T}
+        numerical_dims = dim(a, dims)
+        data = Statistics.$fun(parent(a); dims=numerical_dims, kwargs...)
+        names = symmetric_names(L, numerical_dims)
+        return NamedDimsArray{names}(data)
+    end
+end
+
+function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
+    # 0 Allocations. See `@btime (()-> symmetric_names((:foo, :bar), 1))()`
+    names = if dims == 1
+        (L[2], L[2])
+    elseif dims == 2
+        (L[1], L[1])
+    else
+        (:_, :_)
+    end
+    compile_time_return_hack(names)
+end
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -19,7 +19,7 @@ function matrix_prod_names(a, b)
             "Cannot take matrix product of arrays with different inner dimension names. $a vs $b"
         ))
     res = matmul_names(a, b)
-    compile_time_return_hack(res)
+    return compile_time_return_hack(res)
 end
 
 
@@ -87,5 +87,5 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
     else
         (:_, :_)
     end
-    compile_time_return_hack(names)
+    return compile_time_return_hack(names)
 end
--- a/test/functions_math.jl
+++ b/test/functions_math.jl
@@ -155,14 +155,14 @@ end
     @test inv(nda) * nda ≈ NamedDimsArray{(:b, :b)}([1.0 0; 0 1])
 end
 
-@testset "symmetric results" begin
-    @testset "symmetric_names" begin
-        @test symmetric_names((:a, :b), 1) == (:b, :b)
-        @test symmetric_names((:a, :b), 2) == (:a, :a)
-        @test symmetric_names((:a, :b), 5) == (:_, :_)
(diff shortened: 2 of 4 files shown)


src/functions_math.jl
```diff
@@ -67,3 +67,25 @@ function Base.inv(nda::NamedDimsArray{L, T, 2}) where {L,T}
...
+function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
+    # 0 Allocations. See `@btime (()-> symmetric_names((:foo, :bar), 1))()`
+    names = if dims == 1
+        (L[2], L[2])
+    elseif dims == 2
+        (L[1], L[1])
+    else
+        (:_, :_)
+    end
+    compile_time_return_hack(names)
```
oxinabox:
```suggestion
    return compile_time_return_hack(names)
```

src/functions_math.jl
```diff
@@ -67,3 +67,25 @@ function Base.inv(nda::NamedDimsArray{L, T, 2}) where {L,T}
...
+
+function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
+    # 0 Allocations. See `@btime (()-> symmetric_names((:foo, :bar), 1))()`
+    names = if dims == 1
+        (L[2], L[2])
+    elseif dims == 2
+        (L[1], L[1])
+    else
+        (:_, :_)
+    end
```
oxinabox: Should we generalize this? 

```
function symmetric_names(L::NTuple{N,Symbol} dim::Integer)
    name = dim <= N ? L[dim] : :_
    names = ntuple(i->name, N)
    return compile_time_return_hack(names)
end
```

Do check performance if so

test/functions_math.jl
```diff
@@ -154,3 +154,36 @@ end
...
+        a = rand(4)
+        nda = NamedDimsArray{(:a,)}(a)
+        @test f(nda) isa Number
+        @test f(nda) == f(a)
+    end
+    @testset "high dimensional input" begin
+        @test_throws MethodError f(NamedDimsArray(rand(3, 4, 5), (:a, :b, :c)))
+    end
+end
+@testset "cov corrected=$bool" for bool in (true, false)
```
oxinabox:
```suggestion
@testset "No dim passed, cov corrected=$bool" for bool in (true, false)
```

test/functions_math.jl
```diff
@@ -154,3 +154,36 @@ end
     @test nda * inv(nda) ≈ NamedDimsArray{(:a, :a)}([1.0 0; 0 1])
     @test inv(nda) * nda ≈ NamedDimsArray{(:b, :b)}([1.0 0; 0 1])
 end
+
```
oxinabox: Put all of the below into a `cov/cor` testset

src/functions_math.jl
nickrobinson251: Maybe... Will this ever make sense for the `N != 2` case?

src/functions_math.jl
nickrobinson251: I was being consistent with the style in the file, so will also add a return statement to other places :)

test/functions_math.jl
nickrobinson251: i'll just test the dims case too :)

src/functions_math.jl
nickrobinson251: if we decide to do, then your suggestion also doesn't allocate :)

```
function symmetric_names(L::NTuple{N, Symbol}, dim::Integer) where N
    name = dim <= N ? L[dim] : :_
    names = ntuple(i -> name, N)
    return compile_time_return_hack(names)
end
```

test/functions_math.jl
nickrobinson251: done :)

nickrobinson251:
Shall i bump the patch version, to `0.2.1`?

src/functions_math.jl
oxinabox: I think my solution is a bit cleaner, but your call

nickrobinson251: actually on second look, the suggestion gives the wrong result

`symmetric_names((:a, :b), 1)` should be `(:b, :b)`, but the suggestion gives `(:a, :a)`

oxinabox: Good reason to do it your way

PR #48, opened 2019-06-22
Correct default dims keywords to match Base

### LINKED ISSUE #45: sort doesn't work
```julia
julia> a = [2, 5, 4];

julia> nda = NamedDimsArray(a, :a);

julia> sort(nda)
ERROR: MethodError: no method matching sort!(::Array{Int64,1}; dims=Colon())
```

### LINKED ISSUE #46: Base functions need different `dims` fallbacks
When extending Base functions that take a `dims` keyword [we set the default `dims=:`](https://github.com/invenia/NamedDims.jl/blob/45a358b6a9157bad5a5c683a2e2716d30614c2e6/src/functions.jl#L28), but this does not match Base in all cases. 

These are the cases we get right:

- `sum, prod, maximum, minimum, extrema` 
```julia
julia> using NamedDims, Test

julia> a = rand(3, 3);

julia> nda = NamedDimsArray{(:x, :y)}(a);

julia> for f in (sum, prod, maximum, minimum, extrema)
           @test f(a) == f(nda) == f(a; dims=:)
       end
```

- `mapreduce`
```julia
julia> a = rand(Int, 3, 3);

julia> nda = NamedDimsArray{(:x, :y)}(a);

julia> @test mapreduce(isodd, *, a) == mapreduce(isodd, *, a) == mapreduce(isodd, *, a; dims=:)
```

And these are the cases we need to fix:
- `count` does not take a `dims` keyword
```julia
julia> a = [true false; true true];

julia> count(a)
3

julia> count(a; dims=1)
ERROR: MethodError: no method matching count(::Array{Bool,2}; dims=1)

julia> nda = NamedDimsArray{(:x, :y)}(a);

julia> count(nda)
ERROR: MethodError: no method matching count(::Array{Bool,2}; dims=Colon())
```
- `sort`/`sort!` for multidimensional array does not have a default for the `dims` keyword, and the input must be a `Integer` (#45)
- `cumsum` and `cumprod` are the same as `sort`:  no default, must be `Integer`
-  `mapslices` does not have not have a default for the `dims` keyword, and the input must be a `Integer` OR an iterable of `Integer`s

```julia
julia> a = reshape(Vector(1:8),(2,2,2));

julia> nda = NamedDimsArray{(:x, :y, :z)}(a);

julia> mapslices(sum, a)
ERROR: UndefKeywordError: keyword argument dims not assigned
Stacktrace:
 [1] mapslices(::Function, ::Array{Int64,3}) at ./abstractarray.jl:1919
 [2] top-level scope at none:0

julia> mapslices(sum, a; dims=1)
1×2×2 Array{Int64,3}:
[:, :, 1] =
 3  7

[:, :, 2] =
 11  15

julia> mapslices(sum, a; dims=[1, 3])
1×2×1 Array{Int64,3}:
[:, :, 1] =
 14  22

julia> mapslices(sum, nda)
ERROR: MethodError: no method matching iterate(::Colon)
```

nickrobinson251:
- Closes #45
- Closes #46

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -32,12 +32,12 @@ for (mod, funs) in (
     end
 end
 
-# 1 arg before
+# 1 arg before - no default no `dims` keyword
 for (mod, funs) in (
     (:Base, (:mapslices,)),
 )
     for fun in funs
-        @eval function $mod.$fun(f, a::NamedDimsArray; dims=:, kwargs...)
+        @eval function $mod.$fun(f, a::NamedDimsArray; dims, kwargs...)
             numerical_dims = dim(a, dims)
             data = $mod.$fun(f, parent(a); dims=numerical_dims, kwargs...)
             return nameddimsarray_result(a, data, numerical_dims)
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -39,6 +39,12 @@ for (mod, funs) in (
             data = $mod.$fun(parent(a); dims=numerical_dims, kwargs...)
             return nameddimsarray_result(a, data, numerical_dims)
         end
+
+        # Vector case
+        @eval function $mod.$fun(a::NamedDimsArray{L, T, 1}; kwargs...) where {L, T}
+            data = $mod.$fun(parent(a); kwargs...)
+            return NamedDimsArray{NamedDims.names(a)}(data)
+        end
     end
 end
 
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -17,10 +17,7 @@ end
 
 # 1 Arg
 for (mod, funs) in (
-    (:Base, (
-        :sum, :prod, :maximum, :minimum, :extrema, :cumsum, :cumprod,
-        :sort, :sort!,)
-    ),
+    (:Base, (:sum, :prod, :maximum, :minimum, :extrema)),
     (:Statistics, (:mean, :std, :var, :median)),
 )
     for fun in funs
@@ -32,7 +29,20 @@ for (mod, funs) in (
     end
 end
 
-# 1 arg before - no default no `dims` keyword
+# 1 Arg - no default for `dims` keyword
+for (mod, funs) in (
+    (:Base, (:cumsum, :cumprod, :sort, :sort!)),
+)
+    for fun in funs
+        @eval function $mod.$fun(a::NamedDimsArray; dims, kwargs...)
+            numerical_dims = dim(a, dims)
+            data = $mod.$fun(parent(a); dims=numerical_dims, kwargs...)
+            return nameddimsarray_result(a, data, numerical_dims)
+        end
+    end
+end
+
+# 1 arg before - no default for `dims` keyword
 for (mod, funs) in (
     (:Base, (:mapslices,)),
 )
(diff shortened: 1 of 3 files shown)


src/functions.jl
```diff
@@ -12,16 +12,12 @@ function nameddimsarray_result(original_nda, reduced_data, reduction_dims::Colon
     return reduced_data
 end
 
-
 ###################################################################################
 # Overloads
 
 # 1 Arg
```
oxinabox:
```suggestion
# 1 Arg: reducers 
```

oxinabox: Also rename `nameddimsarray_result` to `nameddimsarray_reduction_result` ?

src/functions.jl
```diff
@@ -33,12 +29,31 @@ for (mod, funs) in (
     end
 end
 
-# 1 arg before
+# 1 Arg - no default for `dims` keyword
+for (mod, funs) in (
```
nicoleepp: For code clarity, might be better to change `funs` and `fun` to `funcs` and `func` which are the more common abbreviation of `functions`? Even if it does makes the code less "fun" :)

src/functions.jl
```diff
@@ -33,12 +29,31 @@ for (mod, funs) in (
...
+    end
+end
+
+# 1 arg before - no default for `dims` keyword
 for (mod, funs) in (
     (:Base, (:mapslices,)),
 )
     for fun in funs
-        @eval function $mod.$fun(f, a::NamedDimsArray; dims=:, kwargs...)
+        @eval function $mod.$fun(f, a::NamedDimsArray; dims, kwargs...)
```
nicoleepp: How does this work without a default for the kwarg `dims`?

test/functions.jl
```diff
@@ -3,43 +3,86 @@ using NamedDims: names
 using Test
 using Statistics
 
-@testset "sum" begin
-    nda = NamedDimsArray([10 20; 30 40], (:x, :y))
+@testset "Base" begin
+a = [10 20; 31 40]
```
nicoleepp: Code within the testset should be indented

src/functions.jl
nickrobinson251: just the same as Base -- without `dims` specified by the user, it is a `UndefKeywordError`

```julia
julia> A = reshape(1:8, (2, 2, 2));

julia> mapslices(sum, A)
ERROR: UndefKeywordError: keyword argument dims not assigned
Stacktrace:
 [1] mapslices(::Function, ::Array{Int64,3}) at ./abstractarray.jl:1919
 [2] top-level scope at none:0
```

test/functions.jl
nickrobinson251: I agree in principle, the problem is that this file was already way over the line length, and indenting it further made writing the code on my little laptop screen almost impossible

I think a better solution is a follow-up PR that fixes up the style in the test - what do you think?

nicoleepp: That seems good to me

src/functions.jl
nickrobinson251: haha 🎊 

For this PR i would prefer to stick with these names as they're already used elsewhere in the file

test/functions.jl
nickrobinson251: Cool - i'll merge this and make a follow up PR for style changes :) thanks!

PR #49, opened 2019-06-22
Fix vector' * vector

### LINKED ISSUE #36: VectorTranspose * Vector broken
```
julia> x = NamedDimsArray{(:a,)}([1,2,3])
3-element NamedDimsArray{(:a,),Int64,1,Array{Int64,1}}:
 1
 2
 3


julia> x'*x
ERROR: MethodError: no method matching NamedDimsArray{(:_,),T,N,A} where A<:AbstractArray{T,N} where N where T(::Int64)
```

Correct answer is:
`14`
Not a `Array` at all, just a scalar.

nickrobinson251:
closes #36

### DIFF
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -42,9 +42,11 @@ end
 function Base.:*(a::NamedDimsArray{L,T,2,<:CoVector}, b::AbstractVector) where {L,T}
     return *(parent(a), b)
 end
+
 function Base.:*(a::CoVector, B::NamedDimsArray{L,T,1,<:AbstractVector}) where {L,T}
     return *(a, parent(b))
 end
+
 """
     @declare_matmul(MatrixT, VectorT=nothing)
 
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -25,7 +25,6 @@ function matrix_prod_names(a, b)
     return compile_time_return_hack(res)
 end
 
-
 for (NA, NB) in ((1,2), (2,1), (2,2))  #Vector * Vector, is not allowed
     @eval function Base.:*(a::NamedDimsArray{A,T,$NA}, b::NamedDimsArray{B,S,$NB}) where {A,B,T,S}
         L = matrix_prod_names(A,B)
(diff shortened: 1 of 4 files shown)


nickrobinson251:
hmm... this isn't as simple as I thought -- i'm not sure `@declare_matmul` works quite as intended

nickrobinson251:
We now have a method ambiguity for `a' * nda` 😞 -- does this seem like the right approach at all or is there another appraoch to fixing this @oxinabox ?

src/functions_math.jl
```diff
@@ -31,6 +33,20 @@ for (NA, NB) in ((1,2), (2,1), (2,2))  #Vector * Vector, is not allowed
...
+    return *(parent(a), parent(b))
+end
+
+function Base.:*(a::NamedDimsArray{L,T,2,<:CoVector}, b::AbstractVector) where {L,T}
+    return *(parent(a), b)
+end
+
+function Base.:*(a::CoVector, B::NamedDimsArray{L,T,1,<:AbstractVector}) where {L,T}
+    return *(a, parent(b))
+end
```
nickrobinson251: If we remove this method, we still get an ambiguity, specifically:
```julia
  Expression: ndv' * ndv == adjoint(v) * ndv == transpose(v) * ndv
  MethodError: *(::Adjoint{Int64,Array{Int64,1}}, ::NamedDimsArray{(:vec,),Int64,1,Array{Int64,1}}) is ambiguous. Candidates:
    *(a::AbstractArray{T,2} where T, b::NamedDimsArray{#s19,#s20,1,A} where A<:AbstractArray{#s20,1} where #s20 where #s19) in NamedDims at /Users/npr/repos/NamedDims.jl/src/functions_math.jl:71
    *(u::Adjoint{T,#s623} where #s623<:(AbstractArray{T,1} where T) where T, v::AbstractArray{T,1} where T) in LinearAlgebra at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/LinearAlgebra/src/adjtrans.jl:197
  Possible fix, define
    *(::Adjoint{T,#s623} where #s623<:(AbstractArray{T,1} where T) where T, ::NamedDimsArray{#s19,#s20,1,A} where A<:AbstractArray{#s20,1} where #s20 where #s19)
```

nickrobinson251: I think this is now resolved... but in the most cumbersome way possible - just list the exactly all of the methods we want - and I don't love it

oxinabox: `@eval` loop?

nickrobinson251: seemed even less nice 🤷‍♂

oxinabox: we use a lot of `@eval` loops

nickrobinson251: okay.  we can write something like, if that seems nicer? :)
```julia
for A in (Adjoint{<:Any, <:AbstractVector}, Transpose{<:Real, <:AbstractVector{<:Real}})
    @eval function Base.:*(a::$A, b::NamedDimsArray{L,T,1,<:AbstractVector{T}}) where {L,T}
        return *(a, parent(b))
    end
end
```

nickrobinson251: changed :)

PR #51, opened 2019-06-24
Update README

nickrobinson251:


README.md
```diff
@@ -12,13 +12,16 @@
...
-For `nda = NamedDimsArray{(:x,:y,:z)}(rand(10,20,30))`.
+For `nda = NamedDimsArray{(:x, :y, :z)}(rand(10, 20, 30))`.
 
  - Unwrapping: `parent(nda)`: returns the underlying `AbstractArray` that is wrapped by the `NamedDimsArray`
- - Indexing: `nda[y=2]`: the same as `nda[x=:, y=2, z=:]` which is the same as `nda[:,2,:]`
+ - Indexing: `nda[y=2]`: the same as `nda[x=:, y=2, z=:]` which is the same as `nda[:, 2, :]`
  - Functions taking a dims arg: `sum(nda; dims=:y)` is the same as `sum(nda; dims=2)`
  - Renaming: `rename(nda, new_names)` returns a new `NamedDimsArray` with the `new_names` but still wrapping the same data.
 
+If names are duplicated, then indexing by name will refer to _both_ dimensions:
```
oxinabox: I'm tempted to say: that this is undefined behavour.
Unless you think we can make this always work?
E.g. for `getindex`

nickrobinson251: Sorry,not sure what you mean.

It is the current behaviour.

Should we open an issue instead?

Or change this comment to a warning: we make no promises about indexing behaviour in the presence of duplicate names and it should not be relied on?

nickrobinson251: What is the current getindex behaviour with duplicate names?

nickrobinson251: (I think i'll ditch that commit and move the duplicate names disucssion to an issue for now)

oxinabox: > Or change this comment to a warning: we make no promises about indexing behaviour in the presence of duplicate names and it should not be relied on

That

nickrobinson251: https://github.com/invenia/NamedDims.jl/issues/52

PR #54, opened 2019-06-25
Make Tracker an optional dependency

### LINKED ISSUE #53: Remove Tracker dependency?
I understand that Tracker is added here so that this package works nicely with Flux + Tracker... and that that is useful... and _maybe_ we can stop supporting Tracker when it stops being the default AutoDiff for Flux...

BUT it's an ugly dependency I do not wish to add to a bunch of non-Flux-using packages where I would like to use NamedDims -- is there a way to avoid the Tracker dependency?

nickrobinson251:
- closes #53 
- as far as i can tell, this loses the compatibility requirement for Tracker and there's no way around that


nickrobinson251:
...apparently an unrelated test started failing -- not sure what the deal there is, but surely unrelated anyway :)

nickrobinson251:
for Tracker compat we can use 
```
if Pkg.installed()["Tracker"] < v"0.2.2"
    ...
end
```
(or something more robust)

but what behaviour do we want in this case?

oxinabox:
```
version = Pkg.installed()["Tracker"]
allowed_versions = Pkg.Types.semver_spec("2.2")
if tracker_version ∈ allowed_versions
    include("tracker_compat.jl")
else
    @warn "Tracker version not compatible with NamedDims. Tracker Compatability functionality has been disabled." version allowed_versions
end
```

nickrobinson251:
> ...apparently an unrelated test started failing -- not sure what the deal there is, but surely unrelated anyway :)

🙀 this is not unrelated!
```
julia> A = rand(3, 3);

julia> count(A; dims=1)
ERROR: function count does not accept keyword arguments
Stacktrace:
 [1] kwfunc(::Any) at ./boot.jl:330
 [2] top-level scope at none:0

julia> using Tracker

julia> count(A; dims=1)
ERROR: MethodError: no method matching count(::Array{Float64,2}; dims=1)
Closest candidates are:
  count(::Any) at reduce.jl:709 got unsupported keyword argument "dims"
  count(::Any, ::SparseArrays.SparseMatrixCSC) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/SparseArrays/src/sparsematrix.jl:77 got unsupported keyword argument "dims"
  count(::Any, ::SparseArrays.SparseVector) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/SparseArrays/src/sparsevector.jl:43 got unsupported keyword argument "dims"
  ...
Stacktrace:
 [1] top-level scope at none:0
```

src/NamedDims.jl
```diff
@@ -3,13 +3,29 @@ using Base: @propagate_inbounds
...
+function include_tracker_compat()
+    tracker_version = Pkg.installed()["Tracker"]
+    allowed_versions = Pkg.Types.semver_spec("0.2.2")
+    if tracker_version ∈ allowed_versions
+        include("tracker_compat.jl")
+    else
+        @warn string(
+            "Tracker version not compatible with NamedDims. ",
+            "Tracker compatability functionality has been disabled."
+        )
```
oxinabox: Why don't you do
```
        @warn string(
            "Tracker version not compatible with NamedDims. ",
            "Tracker compatability functionality has been disabled."
        ) tracker_version allowed_versions
```

nickrobinson251: oops - accidental deletion -- nice catch :)

test/functions.jl
```diff
@@ -98,8 +98,8 @@ nda = NamedDimsArray(a, (:x, :y))
         a = [true false; true true]
         nda = NamedDimsArray(a, (:x, :y))
         @test count(nda) == count(a) == 3
-        @test_throws MethodError count(nda; dims=:x)
-        @test_throws MethodError count(a; dims=1)
+        @test_throws ErrorException count(nda; dims=:x)
```
oxinabox: Comment linking to issue in JuliaLang about this

nickrobinson251:
This again bumps the patch version (to 0.2.4) in case we want to tag a 0.2.3 first

src/NamedDims.jl
```diff
@@ -3,10 +3,31 @@ using Base: @propagate_inbounds
...
 export NamedDimsArray, dim, rename, unname
 
+function __init__()
+    @require Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c" include_tracker_compat()
+end
+
+# For Tracker-NamedDims compatibility we require a Tracker version compatible with 0.2.2.
+# Since we use Requires.jl we have no nicer way to set the Tracker compatability.
+function include_tracker_compat()
+    tracker_version = Pkg.installed()["Tracker"]
```
KristofferC: This is not going to work well. `Pkg.installed()` only looks in the Project.toml but it is not guaranteed that Tracker is there if it is just brought in as a dependency.

KristofferC: Instead, just declare `Tracker` in `[deps]` and set compat on it with `[compat]` like normally. You can then load it conditionally.

nickrobinson251: > declare Tracker in [deps]

this is what we're trying to avoid

nickrobinson251: I do not know how else to find which version of Tracker is _actually_ installed

oxinabox: Oh this is super sad. This will actually break for one of my real used cases too then.
Since i only load `Tracker` via `Flux`

nickrobinson251: ohh, now i see. Gah, silly Pkg not having an easier way to get the installed version of a loaded package. I'll make a PR which ditches this version check (and we hope users are on Tracker v0.2.2, at least for now)

KristofferC: Installing a package (without big binary dependencies) is super cheap. It is loading that is expensive. So what is the problem with having to download Tracker and its dependencies? Most of these will already be downloaded already so will be nops.

oxinabox: Tracker has a bigish binary dependency from SpecialFuns.jl

KristofferC: It's like a 200KB tar download so I would not agree with it being too bigish.

PR #55, opened 2019-06-25
Add function for safely accessing `parent`

nickrobinson251:
- closes #39
- `unname` allow you to ensure you _don't_ have a NamedDimsArray

PR #56, opened 2019-06-25
Change to only increment the patch version

nickrobinson251:
- I forgot that pre v1.0 we can add functionality in patch releases
- #55 bumped the minor version -- this changes that so only the patch version is incremented
- now folks don't have to update compat to get the tiny backward-compatible feature


nickrobinson251:
thanks for suggesting this, @oxinabox !

PR #57, opened 2019-06-25
Stop making Tracker loading depend on version

nickrobinson251:
- No reliable way to identify the version
- More useful to assume a compatible Tracker
- see https://github.com/invenia/NamedDims.jl/pull/54#discussion_r297233900 (Thanks @KristofferC for noticing this)

### DIFF
--- a/src/NamedDims.jl
+++ b/src/NamedDims.jl
@@ -10,22 +10,8 @@ using Statistics
 export NamedDimsArray, dim, rename, unname
 
 function __init__()
-    @require Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c" include_tracker_compat()
-end
-
-# For Tracker-NamedDims compatibility we require a Tracker version compatible with 0.2.2.
-# Since we use Requires.jl we have no nicer way to set the Tracker compatability.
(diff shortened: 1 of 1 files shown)


oxinabox:
Possibly solution in https://github.com/MikeInnes/Requires.jl/issues/3#issuecomment-505496316
but I would rather push that into Requires than do it here.
So lets do this

src/NamedDims.jl
```diff
@@ -10,22 +10,8 @@ using Statistics
...
-    allowed_versions = Pkg.Types.semver_spec("0.2.2")
-    if tracker_version ∈ allowed_versions
-        include("tracker_compat.jl")
-    else
-        @warn string(
-            "Tracker version not compatible with NamedDims. ",
-            "Tracker compatability functionality has been disabled."
-        ) tracker_version allowed_versions
-    end
+    # NOTE: NamedDims only compatible with Tracker v0.2.2; but no nice way to enforce that.
```
oxinabox:
```suggestion
    # NOTE: NamedDims is only compatible with Tracker v0.2.2; but no nice way to enforce that.
```

Issue #34, opened 2019-06-03
Add eachslice


oxinabox:
We should add `eachslice`
in julia 1.1 this can be an overload for `Base.eachslice`
and in earlier we can just export it.

Right now can kinda fake it via; `vec(mapslices(nda; dim=dim))`

Example

```
outputs = NamedDimsArray{(:variates, :observations)}(outputs)
table = mapslices(outputs, dims=:variates) do row
        NamedTuple{colnames, coltypes}(Tuple(row))
end |> vec
```

nickrobinson251:
In 1.1, is this sufficient
```
julia> Base.eachslice(A::NamedDimsArray{L}; dims) where L = eachslice(parent(A); dims=NamedDims.dim(L, dims))

julia> nda = NamedDimsArray(rand(2, 2, 2), (:a, :b, :c));

julia> g = eachslice(nda; dims=:c);

julia> iterate(g)
([0.20502 0.326133; 0.33411 0.0193653], 1)

julia> typeof.(iterate(g))
(SubArray{Float64,2,Array{Float64,3},Tuple{Base.Slice{Base.OneTo{Int64}},Base.Slice{Base.OneTo{Int64}},Int64},true}, Int64)
```
Or do we want to have the `SubArray`s be `NamedDimsArrays` too?

oxinabox:
We want the slice to have names too.
As a rule if a thing can have anamw it should have a name.

But that is not hard, I don't think.
Use `Base.Generator` to do the wrapping.

Trick to me is working out how to do it in 1.0.
Which I think requires fixing Compat.jl

nickrobinson251:
Fixing Compat.jl is the better way to do this...

But if that's loads of work, we could just copy the `Base.eachslice` definition here?

nickrobinson251:
I do not know how to get allocations down on something like this
```julia
function Base.eachslice(a::NamedDimsArray; dims, kwargs...)
     numerical_dims = dim(a, dims)
     data = Base.eachslice(parent(a); dims=numerical_dims, kwargs...)
     new_names = slice_names(a, numerical_dims)
     return Base.Generator(NamedDimsArray{new_names}, data)
end

function slice_names(a::NamedDimsArray{L}, n::Integer) where L
     return compile_time_return_hack((L[1:n-1]..., L[n+1:end]...))
end
```
```julia
julia> a = rand(30, 30, 30);

julia> nda = NamedDimsArray(a, (:a, :b, :c));

[...]

PR #58, opened 2019-07-02
Add `eachslice` for Julia 1.1+

nickrobinson251:
- Partial fix for #34 (1.0 support not added, separate issue now opened for that https://github.com/invenia/NamedDims.jl/issues/59)
- This allocates once more than expected (or maybe twice more, depending on your expectations)
- Joint work with @oxinabox who figured out how to reduce allocations at least this far

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -48,6 +48,19 @@ for (mod, funs) in (
     end
 end
 
+if VERSION > v"1.1-"
+    function Base.eachslice(a::NamedDimsArray{L}; dims, kwargs...) where L
+        numerical_dims = dim(a, dims)
+        slices = eachslice(parent(a); dims=numerical_dims, kwargs...)
+        return Base.Generator(slices) do slice
+            # For unknown reasons (something to do with hoisting?) having this in the
+            # function passed to `Generator` actually results in less memory being allocated
+            names = remaining_dimnames_after_dropping(L, numerical_dims)
+            return NamedDimsArray(slice, names)
+        end
+    end
+end
+
 # 1 arg before - no default for `dims` keyword
 for (mod, funs) in (
     (:Base, (:mapslices,)),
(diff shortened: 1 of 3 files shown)


nickrobinson251:
```julia
julia> a = rand(3, 3, 3);

julia> nda = NamedDimsArray(a, (:a, :b, :c));

julia> @btime (() -> eachslice($a; dims=3))();
  187.253 ns (3 allocations: 64 bytes)

julia> @btime (() -> eachslice($nda; dims=:c))();
  267.420 ns (5 allocations: 112 bytes)

```

test/functions.jl
```diff
@@ -44,6 +44,39 @@ nda = NamedDimsArray(a, (:x, :y))
         end
     end
 
+    @testset "eachslice" begin
+        if VERSION > v"1.1-"
```
oxinabox: Do you think we should delcare `nda` within this testset?
This file is getting a buit long to have to read up to check what it was.

nickrobinson251: Good suggestion -- done!

src/functions.jl
```diff
@@ -48,6 +48,19 @@ for (mod, funs) in (
...
+        slices = eachslice(parent(a); dims=numerical_dims, kwargs...)
+        return Base.Generator(slices) do slice
+            # For unknown reasons (something to do with hoisting?) having this in the
+            # function passed to `Generator` actually results in less memory being allocated
+            names = remaining_dimnames_after_dropping(L, numerical_dims)
+            return NamedDimsArray(slice, names)
+        end
+    end
+end
+
```
oxinabox: Do you  think this is the best file for it ?
If you want to change it then do so, or if not then don't.
(Does not block this PR from being merge)

Everything else here is more programatic.

Maybe it is better in `functions_dims` ?

nickrobinson251: I think it fits with "standard base overloads" more than "messing with dimensions" - and fitted better in the corresponding test file -  but 🤷‍♂

 will leave for now :)

nickrobinson251:
will bump patch version then merge

PR #62, opened 2019-08-22
Some macros

mcabbott:
This adds some convenience macros for adding names to arrays, checking that they have the expected names (either in particular places, or in general), and removing names.


mcabbott:
It occurs to me after writing this (!) that quite a bit of this could be functions. Perhaps `unname(A, (:i,:j))` should call `permutedims` only if necessary. It would be easy to write something like `hasnames(A, (:i, :j))`? Then macros, if they exist, would be very simple conveniences.

PR #64, opened 2019-08-22
PermutedDimsArray etc.

mcabbott:
This treats `PermutedDimsArray` exactly like `permutedims`. 

And it adds a method `unname(A::NamedDimsArray, names)` which is always a view of the parent, with axes ordered to match the given names.

### DIFF
--- a/src/functions_dims.jl
+++ b/src/functions_dims.jl
@@ -57,6 +57,23 @@ for f in (
     end
 end
 
+"""
+    unname(A::NamedDimsArray, names) -> AbstractArray
+
+ Returns the parent array if the given names match those of `A`,
+ otherwise a `transpose` or `PermutedDimsArray` view of the parent data.
+ To ensure a copy, call instead `parent(permutedims(A, names))`.
+"""
+function unname(nda::NamedDimsArray{L,T,N}, names::NTuple{N}) where {L,T,N}
+    perm = dim(nda, names)
+    if perm == ntuple(identity, N)
+        return parent(nda)
+    elseif perm == (2,1)
+        return transpose(parent(nda))
+    else
+        return PermutedDimsArray(parent(nda), perm)
+    end
+end
 
 # reshape
 # For now we only implement the version that drops dimension names
--- a/src/functions_dims.jl
+++ b/src/functions_dims.jl
@@ -21,8 +21,13 @@ end
 function Base.permutedims(nda::NamedDimsArray{L}, perm) where {L}
     numerical_perm = dim(nda, perm)
     new_names = permute_dimnames(L, numerical_perm)
-
     return NamedDimsArray{new_names}(permutedims(parent(nda), numerical_perm))
+end
+
+function Base.PermutedDimsArray(nda::NamedDimsArray{L}, perm) where {L}
+    numerical_perm = dim(nda, perm)
+    new_names = permute_dimnames(L, numerical_perm)
+    return NamedDimsArray{new_names}(PermutedDimsArray(parent(nda), numerical_perm))
 end
 
 for f in (
@@ -36,7 +41,6 @@ for f in (
         new_names = (:_, first(L))
         return NamedDimsArray{new_names}($f(parent(nda)))
     end
-
 
     # Vector Double Transpose
     if f !== :permutedims
--- a/src/functions_dims.jl
+++ b/src/functions_dims.jl
@@ -17,17 +17,15 @@ function Base.dropdims(nda::NamedDimsArray; dims)
     return NamedDimsArray{L}(data)
 end
 
-
-function Base.permutedims(nda::NamedDimsArray{L}, perm) where {L}
-    numerical_perm = dim(nda, perm)
-    new_names = permute_dimnames(L, numerical_perm)
-    return NamedDimsArray{new_names}(permutedims(parent(nda), numerical_perm))
-end
-
-function Base.PermutedDimsArray(nda::NamedDimsArray{L}, perm) where {L}
-    numerical_perm = dim(nda, perm)
-    new_names = permute_dimnames(L, numerical_perm)
-    return NamedDimsArray{new_names}(PermutedDimsArray(parent(nda), numerical_perm))
+for f in (
+    :(Base.permutedims),
+    :(Base.PermutedDimsArray)
+)
+    @eval function $f(nda::NamedDimsArray{L}, perm) where {L}
+        numerical_perm = dim(nda, perm)
+        new_names = permute_dimnames(L, numerical_perm)
+        return NamedDimsArray{new_names}($f(parent(nda), numerical_perm))
+    end
 end
 
 for f in (
@@ -54,24 +52,6 @@ for f in (
     @eval function $f(nda::NamedDimsArray{L,T,2}) where {L,T}
         new_names = (last(L), first(L))
         return NamedDimsArray{new_names}($f(parent(nda)))
-    end
-end
-
-"""
-    unname(A::NamedDimsArray, names) -> AbstractArray
-
- Returns the parent array if the given names match those of `A`,
- otherwise a `transpose` or `PermutedDimsArray` view of the parent data.
- To ensure a copy, call instead `parent(permutedims(A, names))`.
-"""
-function unname(nda::NamedDimsArray{L,T,N}, names::NTuple{N}) where {L,T,N}
-    perm = dim(nda, names)
-    if perm == ntuple(identity, N)
-        return parent(nda)
-    elseif perm == (2,1)
-        return transpose(parent(nda))
-    else
-        return PermutedDimsArray(parent(nda), perm)
     end
 end
 
--- a/test/functions_dims.jl
+++ b/test/functions_dims.jl
@@ -55,7 +55,7 @@ end
     end
 end
 
-@testset "permutedims" begin
+@testset "$permutedims" for permutedims in (permutedims, PermutedDimsArray) begin
     nda = NamedDimsArray{(:w, :x, :y, :z)}(ones(10, 20, 30, 40))
     @test (
         names(permutedims(nda, (:w, :x, :y, :z))) ==
--- a/test/functions_dims.jl
+++ b/test/functions_dims.jl
@@ -55,7 +55,7 @@ end
     end
 end
 
-@testset "$permutedims" for permutedims in (permutedims, PermutedDimsArray) begin
+@testset "$permutedims" for permutedims in (permutedims, PermutedDimsArray)
     nda = NamedDimsArray{(:w, :x, :y, :z)}(ones(10, 20, 30, 40))
     @test (
         names(permutedims(nda, (:w, :x, :y, :z))) ==
--- a/test/functions_dims.jl
+++ b/test/functions_dims.jl
@@ -55,48 +55,37 @@ end
     end
 end
 
-@testset "$permutedims" for permutedims in (permutedims, PermutedDimsArray)
+@testset "$f" for f in (permutedims, PermutedDimsArray)
     nda = NamedDimsArray{(:w, :x, :y, :z)}(ones(10, 20, 30, 40))
     @test (
-        names(permutedims(nda, (:w, :x, :y, :z))) ==
-        names(permutedims(nda, 1:4)) ==
+        names(f(nda, (:w, :x, :y, :z))) ==
+        names(f(nda, 1:4)) ==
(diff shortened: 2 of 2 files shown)


oxinabox:
Can we discuss `unname` in a seperate issue?
I'm much more hestitant about introducing new public API functions,
than about fixing things that already work for `AbstractArrays`.

And there is more to discuss on that one,

src/functions_dims.jl
```diff
@@ -21,10 +21,15 @@ end
...
-
     return NamedDimsArray{new_names}(permutedims(parent(nda), numerical_perm))
 end
 
+function Base.PermutedDimsArray(nda::NamedDimsArray{L}, perm) where {L}
+    numerical_perm = dim(nda, perm)
+    new_names = permute_dimnames(L, numerical_perm)
+    return NamedDimsArray{new_names}(PermutedDimsArray(parent(nda), numerical_perm))
+end
+
```
oxinabox: I lean towards writing this together with `permutedims` as


```
for f in (
    :(Base.permutedims),
    :(Base.PermutedDimsArray)
)
    @eval function $f(nda::NamedDimsArray{L}, perm) where {L}
        numerical_perm = dim(nda, perm)
        new_names = permute_dimnames(L, numerical_perm)
        return NamedDimsArray{new_names}($f(parent(nda), numerical_perm))
    end
end
```

test/functions_dims.jl
```diff
@@ -55,7 +55,7 @@ end
     end
 end
 
-@testset "permutedims" begin
+@testset "$permutedims" for permutedims in (permutedims, PermutedDimsArray)
```
oxinabox: Can we call the function inside this something other than `permutedims` so it doesn't shadow `Base.permutedims`?

Maybe `for permutedim_func in  (permutedims, PermutedDimsArray)`

src/functions_dims.jl
mcabbott: Sure, done!

test/functions_dims.jl
mcabbott: Sure, done! Although I see I went with `f`, same as defn, perhaps less readable...

src/functions_dims.jl
```diff
@@ -21,10 +21,15 @@ end
...
     numerical_perm = dim(nda, perm)
     new_names = permute_dimnames(L, numerical_perm)
-
     return NamedDimsArray{new_names}(permutedims(parent(nda), numerical_perm))
 end
 
+function Base.PermutedDimsArray(nda::NamedDimsArray{L}, perm) where {L}
+    numerical_perm = dim(nda, perm)
+    new_names = permute_dimnames(L, numerical_perm)
+    return NamedDimsArray{new_names}(PermutedDimsArray(parent(nda), numerical_perm))
```
nickrobinson251: Shouldn't this return a `PermutedDimsArray` (wrapping a NamedDimsArray)?

Wouldn't we just want to add a constructor that allows `perm` to contain `Symbol`s?

Since `PermuteDimsArray` (with Ints) already works here, don't we just want:
```julia
julia> nda = NamedDimsArray{(:w, :x, :y, :z)}(ones(10, 20, 30, 40));

julia> pda = PermutedDimsArray(nda, (1, 2, 4, 3));  # already works

julia> typeof(pda)
PermutedDimsArray{Float64,4,(1, 2, 4, 3),(1, 2, 4, 3),NamedDimsArray{(:w, :x, :y, :z),Float64,4,Array{Float64,4}}}

julia> size(pda)
(10, 20, 40, 30)

julia> pda == PermutedDimsArray(nda, (:w, :x, :z, :y));  # this is the method that should be added?
true
```

src/functions_dims.jl
```diff
@@ -53,6 +57,23 @@ for f in (
...
+function unname(nda::NamedDimsArray{L,T,N}, names::NTuple{N}) where {L,T,N}
+    perm = dim(nda, names)
+    if perm == ntuple(identity, N)
+        return parent(nda)
+    elseif perm == (2,1)
+        return transpose(parent(nda))
+    else
+        return PermutedDimsArray(parent(nda), perm)
+    end
+end
```
nickrobinson251: I feel like this does too much? is this behaviour recovered by some combination of the current `unname` and `permutedims`/`PermuteDimsArray`?

mcabbott:
OK, I trimmed this to just `PermutedDimsArray`, with your good suggestions. 

That's interesting about default constructors etc, I didn't realise this was anything beyond a convention. I can't in 5 minutes make `Tracker` produce such errors with tracked `PermutedDimsArray`s, should that be possible?

src/functions_dims.jl
mcabbott: I made #65 now to discuss this. My goal was to guarantee the order of dimensions (for correctness) and not produce things like `Transpose` unless needed (for tidiness).

src/functions_dims.jl
mcabbott: I was hoping `PermutedDimsArray` could return something with names you can refer to. Although @oxinabox raises the issue that it's weird not to return the constructor's type. Perhaps another objection is that if you care about the order of indices, then you are moving to positional not named code anyway, and should unwrap.

oxinabox:
> I can't in 5 minutes make Tracker produce such errors with tracked PermutedDimsArrays, should that be possible?

```
julia> x = [10a + b for a in 1:3, b in 1:4];

julia> trx = TrackedArray(x);

julia> Set(PermutedDimsArray(data) for data in [trx])
ERROR: MethodError: no method matching PermutedDimsArray(::TrackedArray{…,Array{Int64,2}})
```

mcabbott:
But that's just complaining about the lack of a permutation, there's no one-arg method:
```
julia> PermutedDimsArray(x)
ERROR: MethodError: no method matching PermutedDimsArray(::Array{Int64,2})

julia> Set(PermutedDimsArray(data, (2,1)) for data in [trx])
Set(TrackedArray{…,PermutedDimsArray{Int64,2,(2, 1),(2, 1),Array{Int64,2}}}[[11 21 31; 12 22 32; 13 23 33; 14 24 34] (tracked)])
```

src/functions_dims.jl
iamed2: What if we keep `PermutedDimsArray` returning a `PermutedDimsArray`, but add some methods on `PermutedDimsArray{<:NamedDimsArray}` to return/use names? That seems like the alternative option to me. The issue with that approach is that there might be a bunch of methods necessary for dealing with `NamedDimsArray` × `PermutedDimsArray` interactions. Could be prohibitively troublesome.

mcabbott: Couldn't you argue the same for `Transpose`? The only difference is that the function usually called is named `transpose`.

iamed2: Ah right, that might mean that we can just include `PermutedDimsArray` with existing transpose-handling code (with some augmentation).

oxinabox:
Ah, right, yes. This probably can't happen for things that don't have 1arg constructors. Just because of the nature that bug.

src/functions_dims.jl
mcabbott: Which functions do you have in mind? I just discovered that `*` does seem to know about this, `PermutedDimsArray(rand(2,3),(2,1)) * rand(2)` works... but does that cause any surprises?

iamed2: According to @oxinabox this would be a bunch of work so I'll defer to him

nickrobinson251: I dislike this behaviour

nickrobinson251: I think Eric's suggestion is nice, but is too much work and maintainence for the gain

iamed2:
@nickrobinson251's position sounds reasonable to me

src/functions_dims.jl
nickrobinson251: To be clear, I think this is an interesting case... but I find `PermutedDimsArray(::NamedDimsArray, perm) -> NamedDimsArray{PermuteDimsArray}` to be really surprising behaviour. I expect to get a `PermuteDimsArray{NamedDimsArrays}` which is what we get on master. I _do_ think allowing `perm` to be names i.e `Symbol`s make sense.

I'd be happy with special behaviour for `PermuteDimsArray{NamedDimsArrays}` if the maintainence cost is propornate.

Just a personal opinion - open to alternative views :)

mcabbott: It does seem odd that Julia has `transpose` which makes `Transpose` but no `lazypermutedims` which makes `PermutedDimsArray`, so the name of the type does double duty. Maybe that's the core here? This PR wants to extend `lazypermutedims`, although it happens not to have that spelling. 

It would seem very strange to me that functions like `sum(A, dims=:k)` would be expected to work on things not of the form `NamedDimsArray{Storage}`. Anyone writing a function should then have to know to look inside, and not just dispatch on `::NamedDimsArray`. They don't have to do this with `NamedDimsArray{Transpose}` etc.

oxinabox: > It would seem very strange to me that functions like sum(A, dims=:k) would be expected to work on things not of the form NamedDimsArray{Storage}

For the case of wrapper arrays, then unless they are also messing with `dims`, I expect them to delegate that kwarg straight to the inner array type. And if they are, I kind of expect in many cases (Inc `PermuteDimsArray`) they can do their messing about with it and passes them modified version to the wrapped array.
And so it just works,in theory.

Orthogonal features added via wrapper arrays should work no matter the order of the wrapping

mcabbott: > Orthogonal features added via wrapper arrays should work no matter the order of the wrapping

OK, I like this vision of commutative wrappers. But for instance Tracker really needs to be outermost, I think. (I haven't looked at how you made this work.) Maybe it's a hard problem to decide who wins. 

Some functions now seem to dispatch on `::NamedDimsArray`. Perhaps this could be solved by some clever `Union{NDA, Adjoint{NDA}, Transpose{NDA}, ...}`? And some Base functions fail to pass things they see as illegal along, I guess you could also give them extra methods?
```julia
julia> xy = NamedDimsArray{(:x, :y)}(rand(2,2)); 
julia> yx = NamedDimsArray{(:y, :x)}(rand(2,2));

julia> xy * transpose(yx) # correctly errors
ERROR: DimensionMismatch("Cannot take matrix product of arrays with different inner dimension names. (:x, :y) vs (:x, :y)")

julia> xy * PermutedDimsArray(yx, (2,1)) # silently falls back to AbstractArray method
2×2 NamedDimsArray{(:x, :_),Float64,2,Array{Float64,2}}:
 0.571736  0.94319 
 0.379215  0.585958

julia> sum(PermutedDimsArray(yx, (2,1)), dims=:y) # doesn't pass :y through
ERROR: MethodError: no method matching iterate(::Symbol)
```

mcabbott: (I seem to have deleted my comment by accident, sorry)

> Orthogonal features added via wrapper arrays should work no matter the order of the wrapping

I like this vision, but for now it seems many things do care. Some examples:

```julia
julia> xy = NamedDimsArray{(:x, :y)}(rand(2,2)); 

julia> yx = NamedDimsArray{(:y, :x)}(rand(2,2));

julia> xy * transpose(yx) # correctly errors, as NamedDimsArray is outermost
ERROR: DimensionMismatch("Cannot take matrix product of arrays with different inner dimension names. (:x, :y) vs (:x, :y)")

julia> xy * PermutedDimsArray(yx, (2,1)) # silently falls back to AbstractArray
2×2 NamedDimsArray{(:x, :_),Float64,2,Array{Float64,2}}:
 0.571736  0.94319 
 0.379215  0.585958

julia> sum(PermutedDimsArray(yx, (2,1)), dims=:y) # doesn't know how to pass this along
ERROR: MethodError: no method matching iterate(::Symbol)
```
Perhaps one could avoid these by always dispatching on some big `Union{NDA, Adjoint{NDA}, Transpose{NDA}, ...}` type, I don't know how messy that would be. 

But without commutativity of wrapping, I guess deciding who goes outermost might be a hard problem. I suppose my implicit rule here was that `PermutedDimsArray` is just a storage detail, like `Transpose`. User code shouldn't care, so it doesn't need to be exposed.

mcabbott:
Closing in favour of someday making wrappers outside `NamedDimsArray` work smoothly, instead.

mcabbott:
I had a go at making wrappers outside `NamedDimsArray` work, over here: https://github.com/mcabbott/NamedPlus.jl . Some things work, but things like `sum(A; dims=:a)` don't yet look inside the outermost type.

PR #68, opened 2019-10-02
Mutating functions

mcabbott:
This should make `push!` and friends work.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.4"
+version = "0.2.5"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -92,7 +92,7 @@ end
 
 # 1 Arg
 for (mod, funs) in (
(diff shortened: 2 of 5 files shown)


oxinabox:
These look good! 
The `wrapper_array.jl` file is for the minimal (which still a lot) of stuff to implement the `AbstractArray` interface,
the file you want to put them in is probably `functions.jl`

In particular a few ot them `push!` etc can be added to the `@eval` loop here
https://github.com/invenia/NamedDims.jl/blob/217f4d2d5e74f0a564d5096cf4d9af53470f40ff/src/functions.jl#L95

If you're keen you could also do `pop!` and `popfirst!`
but if not then that is fine.

A few general (minor) comments:
 - rather than using `nda.data` we use `parent(nda)`, and consistency is good.
 - We're now offically using the [Blue style-guide](https://github.com/invenia/BlueStyle), which mandates that all multiline functions explictly use `return`.
 - I am a big fan of nesting testsets, to keep it clear what is testing what.
 - Can more of the tests directly  the array values after mutating? Rather than testing things like `length`
```
push!(nda, 2)
@test parent(nda) == [1, 2, 3, 2]
```

mcabbott:
Will do. BTW the reason I wrote things like `@test length(push!(ndv, 40)) == 4` is to put the function being tested within `@test`.

oxinabox:
> Will do. BTW the reason I wrote things like @test length(push!(ndv, 40)) == 4 is to put the function being tested within @test.

I use lots of small nested testsets for the same reason

PR #72, opened 2019-10-03
sum!, prod! etc

mcabbott:
These give an error if names don’t match. I guess the other option would be to permute things to make it work if possible.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -114,8 +114,14 @@ for (mod, funs) in (
                 return NamedDimsArray{L}(data)
             end
 
+            function $mod.$fun(a::AbstractArray, b::NamedDimsArray{L}) where L
+                data = $mod.$fun(a, parent(b))
+                newL = L[1:ndims(a)]
+                return NamedDimsArray{newL}(data)
+            end
+
             function $mod.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
-                newL = unify_names_longest(La, Lb)
+                newL = unify_names_longest(La, Lb)[1:ndims(a)]
                 data = $mod.$fun(parent(a), parent(b))
                 return NamedDimsArray{newL}(data)
             end
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -115,9 +115,9 @@ for (mod, funs) in (
             end
 
             function $mod.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
-                unify_names_longest(La, Lb)
+                newL = unify_names_longest(La, Lb)
                 data = $mod.$fun(parent(a), parent(b))
-                return NamedDimsArray{La}(data)
+                return NamedDimsArray{newL}(data)
             end
 
         end
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -110,18 +110,20 @@ for (mod, funs) in (
         @eval begin
 
             function $mod.$fun(a::NamedDimsArray{L}, b::AbstractArray) where L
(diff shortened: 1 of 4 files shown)


src/functions.jl
```diff
@@ -102,6 +102,28 @@ for (mod, funs) in (
...
+    for fun in funs
+        @eval begin
+
+            function $mod.$fun(a::NamedDimsArray{L}, b::AbstractArray) where L
+                data = $mod.$fun(parent(a), b)
+                return NamedDimsArray{L}(data)
+            end
+
+            function $mod.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
+                newL = unify_names_longest(La, Lb)
```
oxinabox: why `unify_names_longest` and not just `unify_names`?

test/functions.jl
```diff
@@ -44,6 +44,18 @@ using Statistics
         end
     end
 
+    @testset "$f!" for (f,f!) in zip((sum, prod, maximum, minimum), (sum!, prod!, maximum!, minimum!))
+        nda1 = sum(nda, dims=1)
+        nda2 = sum(nda, dims=2)
```
oxinabox: Can it be made more clear in this test when we are testing a column vector vs row vector etc.
I think at least it is woth declaring all things inside this testset, 
rather than using ones declared earlier.

(I've been tranistioning more testsets over to being self contained like that )

src/functions.jl
mcabbott: Because this works: `sum!(zeros(2), rand(2,10))`. 

However it's still not quite right, as I want only `newL[1:ndims(a)]`.

src/functions.jl
```diff
@@ -102,6 +102,28 @@ for (mod, funs) in (
...
+                data = $mod.$fun(parent(a), parent(b))
+                return NamedDimsArray{newL}(data)
+            end
+
```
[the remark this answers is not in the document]
mcabbott: As I added an extra indent level.

src/functions.jl
```diff
@@ -114,8 +114,14 @@ for (mod, funs) in (
...
 
+            function $mod.$fun(a::AbstractArray, b::NamedDimsArray{L}) where L
+                data = $mod.$fun(a, parent(b))
+                newL = L[1:ndims(a)]
+                return NamedDimsArray{newL}(data)
+            end
+
             function $mod.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
-                newL = unify_names_longest(La, Lb)
+                newL = unify_names_longest(La, Lb)[1:ndims(a)]
```
oxinabox: Does that optimize away?

mcabbott: Maybe not, sadly?
```
m = rand(2,2);
@btime (m -> (1,2,3,4)[1:ndims(m)])($m) # 274.245 ns (2 allocations: 128 bytes)
```

mcabbott: Maybe I give up and just return `a`, without filling in any missing names it may have.

oxinabox: Maybe just adapt the `unify_named_longest` 
to be either `unify_named_shortest` (will it error if the first one isn't shortest anyway?)
or `unify_names_first`

That code is in `name_core.jl`

mcabbott: Good idea, done. 

Notice that this means we don’t exactly return the mutated `a`. But still another wrapper of the same data.

oxinabox: > Notice that this means we don’t exactly return the mutated a. But still another wrapper of the same data.

That is inline I feel with how it should be,
getting rid of wild-card names where possible

PR #73, opened 2019-10-10
eachslice + collect

### LINKED ISSUE #59: Support `eachslice` for Julia 1.0
`Base.eachslice` was added in Julia 1.1 (and added to NamedDims by #58)

But we may also want to implement this function for 1.0 and export it
(the current NamedDims method calls `Base.eachslice` internally so obviously does not work for 1.0)

Also see #34

mcabbott:
I believe this closes #59 and #34.

I ‘m not sure why write a custom generator for `eachslice`. This PR changes it to just looks up `dims`, then run the same code as base:
```julia
julia> ab = NamedDimsArray((1:3) .+ zeros(Int,3)', (:a, :b));

julia> @btime eachslice($ab, dims=:b); # master: 319.215 ns (5 allocations: 112 bytes)
  220.690 ns (3 allocations: 64 bytes) 

julia> @btime eachslice($(parent(ab)), dims=2);
  216.897 ns (3 allocations: 64 bytes)

julia> @btime [sum(x)^2 for x in eachslice($ab, dims=:b)]; # master: 14.636 μs (40 allocations: 1.73 KiB)
  383.851 ns (8 allocations: 336 bytes)
```
Still much slower than `eachcol`, for which there is no special code at all:
```julia
julia> NamedDims.names(first(eachcol(ab)))
(:a,)

julia> @btime eachcol($ab);
  12.992 ns (2 allocations: 48 bytes)

julia> @btime [sum(x)^2 for x in eachcol($ab)];
  76.208 ns (7 allocations: 320 bytes)
```

It also adds the simplest possible overload of `collect` so that simple comprehensions will keep names. Doesn’t work on generators of generators, nor products:
```julia
julia> [x^2 for x in ab]
3×3 NamedDimsArray{(:a, :b),Int64,2,Array{Int64,2}}:
 1  1  1
 4  4  4
 9  9  9

julia> NamedDims.names([sum(x)^2 for x in eachcol(ab)]) # doesn't work yet
(:_,)
```

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -12,7 +12,7 @@ function nameddimsarray_result(original_nda, reduced_data, reduction_dims::Colon
     return reduced_data
 end
 
-###################################################################################
+################################################
 # Overloads
 
 # 1 Arg
@@ -44,19 +44,6 @@ for (mod, funs) in (
         @eval function $mod.$fun(a::NamedDimsArray{L, T, 1}; kwargs...) where {L, T}
             data = $mod.$fun(parent(a); kwargs...)
             return NamedDimsArray{NamedDims.names(a)}(data)
-        end
-    end
-end
-
-if VERSION > v"1.1-"
-    function Base.eachslice(a::NamedDimsArray{L}; dims, kwargs...) where L
-        numerical_dims = dim(a, dims)
-        slices = eachslice(parent(a); dims=numerical_dims, kwargs...)
-        return Base.Generator(slices) do slice
-            # For unknown reasons (something to do with hoisting?) having this in the
-            # function passed to `Generator` actually results in less memory being allocated
-            names = remaining_dimnames_after_dropping(L, numerical_dims)
-            return NamedDimsArray(slice, names)
         end
     end
 end
@@ -109,4 +96,34 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
     newL = unify_names(L, names(B))
     data = append!(parent(A), unname(B))
     return NamedDimsArray{newL}(data)
+end
+
+################################################
+# Generators
+
+if VERSION > v"1.1-"
+    Base.eachslice(A::NamedDimsArray; dims) = _eachslice(A, dims)
+else
+    eachcol(A::AbstractVecOrMat) = (view(A, :, i) for i in axes(A, 2))
+    eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
+    # every line identical to Base, but no _eachslice(A, dims) to disatch on.
+    eachslice(A::AbstractArray; dims) = _eachslice(A, dims)
+end
+
+function _eachslice(A::AbstractArray, dims::Symbol)
+    numerical_dims = dim(A, dims)
+    return _eachslice(A, numerical_dims)
+end
+function _eachslice(A::AbstractArray, dims::Tuple)
+    length(dims) == 1 || throw(ArgumentError("only single dimensions are supported"))
+    return _eachslice(A, first(dims))
+end
+@inline function _eachslice(A::AbstractArray, dim::Int)
+    dim <= ndims(A) || throw(DimensionMismatch("A doesn't have $dim dimensions"))
+    idx1, idx2 = ntuple(d->(:), dim-1), ntuple(d->(:), ndims(A)-dim)
+    return (view(A, idx1..., i, idx2...) for i in axes(A, dim))
+end
+
+function Base.collect(itr::Base.Generator{<:NamedDimsArray{L}}) where {L}
+    NamedDimsArray{L}(collect(Base.Generator(itr.f, parent(itr.iter))))
 end
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -101,7 +101,7 @@ end
 ################################################
 # Generators
 
-if VERSION > v"1.1-"
+@static if VERSION >= v"1.1"
     Base.eachslice(A::NamedDimsArray; dims) = _eachslice(A, dims)
 else
     eachcol(A::AbstractVecOrMat) = (view(A, :, i) for i in axes(A, 2))
--- a/test/functions.jl
+++ b/test/functions.jl
@@ -45,7 +45,9 @@ using Statistics
     end
 
     @testset "eachslice" begin
-        if VERSION > v"1.1-"
+        if VERSION < v"1.1-"
+            using NamedDims: eachslice
+        end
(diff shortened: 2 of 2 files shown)


src/functions.jl
```diff
@@ -110,3 +97,33 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
     data = append!(parent(A), unname(B))
     return NamedDimsArray{newL}(data)
 end
+
+################################################
+# Generators
+
+if VERSION > v"1.1-"
```
nickrobinson251:
```suggestion
@static if VERSION > v"1.1-"
```

src/functions.jl
```diff
@@ -110,3 +97,33 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    Base.eachslice(A::NamedDimsArray; dims) = _eachslice(A, dims)
+else
+    eachcol(A::AbstractVecOrMat) = (view(A, :, i) for i in axes(A, 2))
+    eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
+    # every line identical to Base, but no _eachslice(A, dims) to disatch on.
```
nickrobinson251: @oxinabox previously objected to copying Base code here, so I'll defer to him on this

test/functions.jl
```diff
@@ -67,7 +69,29 @@ using Statistics
...
+        if VERSION < v"1.1-"
+            using NamedDims: eachrow, eachcol
         end
+        nda = NamedDimsArray([10 20; 31 40], (:x, :y))
+
+        @test names(first(eachcol(nda))) == (:x,)
+        @test names(first(eachrow(nda))) == (:y,)
+
+        @test_broken names(collect(eachcol(nda))) == (:y,)
+        @test_broken names(collect(eachrow(nda))) == (:y,)
```
nickrobinson251: why are these `test_broken`?
If there's a good reason, let's open an issue to fix them, and comment a link to that issue here

test/functions.jl
```diff
@@ -67,7 +69,29 @@ using Statistics
...
+    end
+
+    @testset "collect" begin
+        ndv = NamedDimsArray([1, 9, 7, 3], :vec)
+        @test names([sqrt(x) for x in ndv]) == (:vec,)
+
+        nda = NamedDimsArray([10 20; 31 40], (:x, :y))
+        @test names([x^2 for x in nda]) == (:x, :y)
+
+        @test_broken names([x^2 + y for x in nda, y in ndv]) == (:x, :y, :vec)
```
nickrobinson251: same as above

test/functions.jl
oxinabox: This doesn't seem like it is broken at all.
that is a normal vector of named vectors.

oxinabox:
>  It also adds the simplest possible overload of collect so that simple comprehensions will keep names. Doesn’t work on generators of generators, nor products:

Can this be in a seperare PR, it is a more contentious feature that IDK how I feel about right now

src/functions.jl
oxinabox:
```suggestion
@static if VERSION >= v"1.1"
```
No need to worry about prerelease versions

src/functions.jl
```diff
@@ -110,3 +97,33 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    return _eachslice(A, numerical_dims)
+end
+function _eachslice(A::AbstractArray, dims::Tuple)
+    length(dims) == 1 || throw(ArgumentError("only single dimensions are supported"))
+    return _eachslice(A, first(dims))
+end
+@inline function _eachslice(A::AbstractArray, dim::Int)
+    dim <= ndims(A) || throw(DimensionMismatch("A doesn't have $dim dimensions"))
+    idx1, idx2 = ntuple(d->(:), dim-1), ntuple(d->(:), ndims(A)-dim)
+    return (view(A, idx1..., i, idx2...) for i in axes(A, dim))
```
oxinabox: Is this how Base does it?
seems crazy, but might compile away

src/functions.jl
```diff
@@ -110,3 +97,33 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    numerical_dims = dim(A, dims)
+    return _eachslice(A, numerical_dims)
+end
+function _eachslice(A::AbstractArray, dims::Tuple)
+    length(dims) == 1 || throw(ArgumentError("only single dimensions are supported"))
+    return _eachslice(A, first(dims))
+end
+@inline function _eachslice(A::AbstractArray, dim::Int)
+    dim <= ndims(A) || throw(DimensionMismatch("A doesn't have $dim dimensions"))
+    idx1, idx2 = ntuple(d->(:), dim-1), ntuple(d->(:), ndims(A)-dim)
```
oxinabox: This shoujld be on two seperate lines
I would title it `inds_before` and `inds_after`

mcabbott: While I agree that that would be clearer, on its own, I think staying line-for-line identical to Base is a good idea if copying.

test/functions.jl
mcabbott: I think the name of the generator dimension would ideally be passed along, as is now done for `[x^s for x in nda]`. But I couldn't see a way to make this work without https://github.com/JuliaLang/julia/pull/32310 .

src/functions.jl
mcabbott: Other approaches are possible, I see that Compat.jl does have eachslice etc. I don't know what you think of depending on that.

test/functions.jl
mcabbott: No, different -- "Doesn’t work on generators of generators, nor products" this is the 2nd case. But now removed if you think this is an awful way of marking such things in code.

mcabbott:
Perhaps we should have an issue for how many more general generators can or should be handled? Making `[sqrt(x) for x in ab]` behave like `map(sqrt, ab)` seemed straightforward.

src/functions.jl
nickrobinson251: > I see that Compat.jl does have eachslice etc. I don't know what you think of depending on that.

Now that Compat.jl is for Julia v1.0+ I'd be happy to use it, rather than copy code here :)

mcabbott:
> Now that Compat.jl is for Julia v1.0+ I'd be happy to use it, rather than copy code here :)

OK then perhaps I close this, and leave you to make a Compat PR. 

>  it is a more contentious feature ...

If you believe `map` & `collect`ed generators ought not to simply propagate names along, then perhaps open a discussion issue?

PR #74, opened 2019-10-14
selectdim

mcabbott:
This makes things like `selectdim(A, :a, 3)` look up index `:a`.


oxinabox:
Looks good to me.
Can you bump the patch version number so I can tag a release?

PR #76, opened 2019-10-30
Allocation tests from comments

mcabbott:
This copies all the `@btime` comments in source to be real tests. 

Motivated by noticing that the one for `order_named_inds` fails on Julia 1.2, which leads to slow indexing:
```
data = rand(2,3);
nda = NamedDimsArray(data, (:row, :col))

@btime $data[2,3]
@btime $nda[2,3]  # good, 1 ns
@btime $nda[row=2, col=3] # bad, 200 ns (or 50 ns on 1.1)
```
I have a partial fix (50ns, but still nonzero allocations) which I'll add if you don't object, just two lines.

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -118,8 +118,9 @@ Returns the values of the `named_inds`, sorted as per the order they appear in `
 with any missing dimnames, having there value set to `:`.
 An error is thrown if any dimnames are given in `named_inds` that do not occur in `dimnames`.
 """
-function order_named_inds(dimnames::Tuple{Vararg{Symbol,N}}; named_inds...) where {N}
-    # 0-Allocations: see `@code_typed (()->order_named_inds((:a, :b, :c), (b=1, c=2)))()`
+order_named_inds(dimnames::Tuple; named_inds...) = order_named_inds(dimnames, named_inds.data)
+
+function order_named_inds(dimnames::Tuple{Vararg{Symbol,N}}, named_inds::NamedTuple) where {N}
     if !tuple_issubset(keys(named_inds), dimnames)
         throw(DimensionMismatch("Expected subset of $(dimnames), got $(keys(named_inds))"))
     end
--- a/test/name_core.jl
+++ b/test/name_core.jl
@@ -4,6 +4,8 @@ using NamedDims:
     unify_names,
     unify_names_longest,
     unify_names_shortest,
+    dim_noerror,
+    tuple_issubset,
     order_named_inds,
     permute_dimnames,
     remaining_dimnames_from_indexing,
@@ -23,6 +25,9 @@ using Test
         @test dim((:x, :y, :a, :b, :c, :d), :d) == 6
         @test_throws ArgumentError dim((:x, :y, :a, :b, :c, :d), :z) # not found
     end
+    @test 0 == @allocated (()->dim((:a, :b), :b))()
+    @test 0 == @allocated (()->dim((:a,:b), (:a,:b)))()
+    @test 0 == @allocated (()->dim_noerror((:a, :b, :c), :c))()
 end
 
  @testset "unify_names_*" begin
@@ -51,7 +56,11 @@ end
         @test_throws DimensionMismatch unify((:a,), (:b,))
         @test_throws DimensionMismatch unify((:a,:b), (:b, :a))
         @test_throws DimensionMismatch unify((:a, :b, :c), (:_, :_, :d))
+
+        @test 0 == @allocated (()->unify((:a, :b), (:a, :_)))()
     end
+    @test 0 == @allocated (()->unify_names_longest((:a, :b), (:a, :_, :c)))()
+    @test 0 == @allocated (()->unify_names_shortest((:a, :b), (:a, :_, :c)))()
 end
 
 @testset "order_named_inds" begin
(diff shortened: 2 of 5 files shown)


oxinabox:
This rocks!
Awesome work.

Can you put them in their own little sub-testsets?

mcabbott:
Should pass now. 

Will making more groups make it more readable? e.g. grouping all the `@test_throws` would add clutter, which I presume is why you didn't.

Partial fix BTW is to write this (which I wonder if you once had, going by the comments):
```
order_named_inds(dimnames::Tuple; named_inds...) = order_named_inds(dimnames, named_inds.data)
function order_named_inds(dimnames::Tuple{Vararg{Symbol,N}}, named_inds::NamedTuple) where {N}
    # as before
```

test/name_core.jl
```diff
@@ -63,6 +85,13 @@ end
     @test order_named_inds((:x, :y); y=2, ) == (:, 2)
     @test order_named_inds((:x, :y); y=20, x=30) == (30, 20)
     @test order_named_inds((:x, :y); x=30, y=20) == (30, 20)
+
+    @test_broken 0 == @allocated (()->order_named_inds((:a, :b, :c), (b=1, c=2)))() # from code comment
+    if v"1.1-" <= VERSION <= v"1.1.99" # test passes on 1.1, but travis has 1.1.1
```
oxinabox: Does this work?
```suggestion
    if v"1.1-" <= VERSION <= v"1.2-" # test passes on 1.1, but travis has 1.1.1
```

mcabbott: Didn't try it, don't have 1.1.1 locally

oxinabox:
> Will making more groups make it more readable? e.g. grouping all the @test_throws would add clutter, which I presume is why you didn't.

I think it would help because it would give more info.
Could left them up a level so it goes:

```
@testset "foo" begin
...
end

@testset "foo allocs" begin
...
end
```


> Partial fix BTW is to write this (which I wonder if you once had, going by the comments), shall I add it?

I think I would need to see it in the context make the commit, and then if it seems bad can always `git revert` that commit out?

mcabbott:
OK have added this partial fix, and its adjustment of tests. 

And maybe I see that you mean groups for the sake of prettier printing of test results, while I was thinking readable source. Which is true, but I'm a little lazy to sort out `for unify in ...` etc nicely and I must run now, sorry. I presume you have rights to add a commit to this branch?

src/name_core.jl
```diff
@@ -118,8 +118,9 @@ Returns the values of the `named_inds`, sorted as per the order they appear in `
 with any missing dimnames, having there value set to `:`.
 An error is thrown if any dimnames are given in `named_inds` that do not occur in `dimnames`.
 """
-function order_named_inds(dimnames::Tuple{Vararg{Symbol,N}}; named_inds...) where {N}
-    # 0-Allocations: see `@code_typed (()->order_named_inds((:a, :b, :c), (b=1, c=2)))()`
+order_named_inds(dimnames::Tuple; named_inds...) = order_named_inds(dimnames, named_inds.data)
```
oxinabox: Can we just get rid of this method entirely,
and demand that the `named_inds` get passed as a NamedTuple 
(or maybe also allow a `Base.Iterators.Pairs`)

Would also need to get rid of the one here:
https://github.com/invenia/NamedDims.jl/blob/f8f77a5119e589eae6f70e548836f7262767b263/src/wrapper_array.jl#L160

and change this call 
https://github.com/invenia/NamedDims.jl/blob/f8f77a5119e589eae6f70e548836f7262767b263/src/wrapper_array.jl#L169
and this call
https://github.com/invenia/NamedDims.jl/blob/f8f77a5119e589eae6f70e548836f7262767b263/src/wrapper_array.jl#L195

oxinabox:
You know what the other great part about having these tests is?
They will prevent the language changing in ways that break this.
Because of PkgEval
(or at least will prompt a core dev to come put in the fix in this package)

src/name_core.jl
mcabbott: You could, I tried some such things, and didn't find a way to get to zero allocations & full speed. So this was the minimal way to change things. It's a clear improvement, and if I find a complete solution I'll be sure to make a PR.

oxinabox: Thats fair.
I  am happy to merge the PR with this still in it

test/wrapper_array.jl
```diff
@@ -167,3 +169,29 @@ end
...
+
+    @test 0 == @allocated NamedDimsArray(cnda, (:x, :y))
+    if VERSION >= v"1.1"
+        @test 0 == @allocated NamedDimsArray(cnda, (:x, :_))
+    else
+        @test_broken 0 == @allocated NamedDimsArray(cnda, (:x, :_))
+    end
+
+    # indexing
+    @test 0 == @allocated cnda[1,1]
```
oxinabox: spacing here: `cnda[1, 1]`

oxinabox:
Looking good.
How do you feel about refactoring it so that rather than repeating so much we have a common macro to do this test?
SOmething like
```
macro test_is_constant_folded(expr)
     quote
         @test 0 == @allocated((()->$(esc(expr)))())
         @inferred ((()->$(esc(expr)))())
         # other tests we might add later
     end
end

If you are not interested then can make a issue and it can happen later.

oxinabox:
> But I think I'm done with this for now.

So you are ready for this to be merged?

mcabbott:
Yes that’s what I meant, sorry.

PR #77, opened 2019-11-03
names(permutedims(transpose(ndv)))

mcabbott:
Minimal fix of a bug that crept in.

Maximal fix is this:
```
    @eval function $f(nda::NamedDimsArray{L}) where (L)
        data = $f(parent(nda))
        new_names = if ndims(nda) == 1 # vector input
            (:_, first(L))
        elseif ndims(data) == 1 # vector output
            (last(L),)
        else
            (last(L), first(L))
        end
        return NamedDimsArray(data, new_names)
    end
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.8"
+version = "0.2.9"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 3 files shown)


oxinabox:
What would you say the advantages/disadvantages of the maximal fix would be?

mcabbott:
It's more compact? Ease of understanding is in the eye of the beholder, but perhaps this requires less thinking about what `permutedims` will produce? 

I was wondering whether there are any exotic types on which the behaviour may be different, but don't know of any. For example Strided.jl makes permutedims lazy... but seems to be careful not to alter what `ndims` is produced:
```
julia> v = [1,2,3];

julia> vsp = @strided permutedims(v)
1×3 reshape(::StridedView{Int64,1,Array{Int64,1},typeof(identity)}, 1, 3) with eltype Int64:
 1  2  3

julia> permutedims(vsp)
3×1 StridedView{Int64,2,Array{Int64,1},typeof(identity)}:
 1
 2
 3
```
I haven't tried to stress it, to see whether say 1.0 may fail to infer through `ndims(data)==1` or something.

PR #79, opened 2019-11-05
isequal and ==

mcabbott:
This declares that arrays with incompatible names are not equal.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -90,6 +90,13 @@ end
 ################################################
 # Non-dim Overloads
 
+for fun in (:(==), :isequal)
+    @eval function Base.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
+        names_are_unifiable(La, Lb) || return false
+        return $fun(parent(a), parent(b))
+    end
+end
+
 # Array then perhaps other args
 for (mod, funs) in (
     (:Base, (:zero, :one, :copy, :empty!, :push!, :pushfirst!)),
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -90,10 +90,10 @@ end
 ################################################
 # Non-dim Overloads
 
-for fun in (:(==), :isequal)
-    @eval function Base.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}) where {La, Lb}
+for fun in (:(==), :isequal, :isapprox)
+    @eval function Base.$fun(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}; kw...) where {La, Lb}
         names_are_unifiable(La, Lb) || return false
-        return $fun(parent(a), parent(b))
+        return $fun(parent(a), parent(b); kw...)
     end
 end
 
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -162,13 +162,14 @@ This is the type of name combination used for binary array operations.
 Where the dimensions of both arrays must be the same.
 """
 function unify_names(names_a, names_b)
+    # @btime (()->unify_names((:a, :b), (:a, :_)))()
     names_a === names_b && return names_a
     length(names_a) == length(names_b) || incompatible_dimension_error(names_a, names_b)
 
     ret = try_unify_names(names_a, names_b)
     ret isa Tuple{Vararg{Symbol}} || incompatible_dimension_error(names_a, names_b)
 
-    return ret
+    return compile_time_return_hack(ret)
 end
 
 function try_unify_names(names_a, names_b)
@@ -180,7 +181,7 @@ function try_unify_names(names_a, names_b)
         a === b && return a
         return false  # mismatch occured, we mark this with a non-Symbol result
     end
-    return ret # can't use compile_time_return_hack but not needed
+    return ret
 end
 
 function names_are_unifiable(names_a, names_b)
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -240,6 +240,26 @@ function unify_names_shortest(names_a, names_b)
     return compile_time_return_hack(ret)
 end
 
+# based on unify_names, but returns true/false
+function names_are_unifiable(names_a, names_b)
+    # @btime (()->names_are_unifiable((:a, :b), (:a, :b)))()
+    # @btime (()->names_are_unifiable((:a, :b), (:a, :_)))()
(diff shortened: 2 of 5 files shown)


src/name_core.jl
```diff
@@ -162,26 +162,31 @@ This is the type of name combination used for binary array operations.
...
         a === :_ && return b
         b === :_ && return a
         a === b && return a
-
-        return false  # mismatch occured, we mark this with a nonSymbol result
+        return false  # mismatch occured, we mark this with a non-Symbol result
     end
-    ret isa Tuple{Vararg{Symbol}} || incompatible_dimension_error(names_a, names_b)
-    return compile_time_return_hack(ret)
+    return ret # can't use compile_time_return_hack but not needed
```
oxinabox: should we do 
```
return ret isa Tuple{Vararg{Symbol}} ? ret : nothing
```

to be inline with the convention for `try_` functions?
I think it will optimize out.

src/functions.jl
```diff
@@ -90,6 +90,13 @@ end
 ################################################
 # Non-dim Overloads
 
+for fun in (:(==), :isequal)
```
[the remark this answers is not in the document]
oxinabox: hmmm, ok.
Should a do `isapprox` here too?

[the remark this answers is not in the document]
mcabbott: I hadn’t tried, but in fact it already works, via broadcasting:
```
julia> nda ≈ nda2
true

julia> nda ≈ nda3
ERROR: DimensionMismatch("Incompatible dimension names: (:x, :y) ≠ (:x, :z)")
```
Can still add if you like. Or just add to tests?

src/name_core.jl
mcabbott: Can do, that’s probably better than passing around the tuple with false in it.

src/functions.jl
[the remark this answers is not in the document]
mcabbott: Sorry that’s no good, we want `false` not an error.

oxinabox:
will review on monday

mcabbott:
Clever not to say which monday! :)

oxinabox:
I will review on monday.
Sorry It slipped under the run and I had a ton of things going on from start of december til start of last week

src/name_core.jl
```diff
@@ -162,26 +162,36 @@ This is the type of name combination used for binary array operations.
 Where the dimensions of both arrays must be the same.
 """
 function unify_names(names_a, names_b)
-    # 0-Allocations if inputs are the same
-    # 0-Allocation, if has a `:_` see  `@btime (()->unify_names((:a, :b), (:a, :_)))()`
-
+    # @btime (()->unify_names((:a, :b), (:a, :_)))()
     names_a === names_b && return names_a
+    length(names_a) == length(names_b) || incompatible_dimension_error(names_a, names_b)
```
oxinabox: Should  this line be in `try_unify_names`  ?
(returning `nothing` rather than erroring)

src/name_core.jl
```diff
@@ -162,26 +162,36 @@ This is the type of name combination used for binary array operations.
 Where the dimensions of both arrays must be the same.
 """
 function unify_names(names_a, names_b)
-    # 0-Allocations if inputs are the same
-    # 0-Allocation, if has a `:_` see  `@btime (()->unify_names((:a, :b), (:a, :_)))()`
-
+    # @btime (()->unify_names((:a, :b), (:a, :_)))()
     names_a === names_b && return names_a
```
oxinabox: can this line be moved into `try_unify_names` ?
It should constant fold alway, right?

src/name_core.jl
```diff
@@ -162,26 +162,36 @@ This is the type of name combination used for binary array operations.
...
+    if ret isa Tuple{Vararg{Symbol}}
+        return compile_time_return_hack(ret)
+    else
+        return nothing
+    end
+end
+
+function names_are_unifiable(names_a, names_b)
+    names_a === names_b && return true
+    length(names_a) == length(names_b) || return false
```
oxinabox: if we moved the above two into `try_unify_names`  then we coujld delete the above two lines

test/functions.jl
```diff
@@ -260,6 +260,23 @@ using Statistics
...
+
+        @testset "$eq" for eq in (Base.:(==), isequal, isapprox)
+            @test eq(nda, nda)
+            @test eq(nda, nda2)
+            @test eq(nda, nda3) == false
+            @test eq(nda, nda4) == false
+            @test eq(nda, ndv) == false
+        end
+        @test isapprox(nda, nda4; atol=2π)
+    end
```
oxinabox: Do we want to add an allocation test here?

src/name_core.jl
```diff
@@ -160,34 +163,48 @@ For example:
...
-    names_a === names_b && return names_a
+names_are_unifiable(names_a, names_b) = try_unify_names(names_a, names_b) !== nothing
 
-    # Error message should not include names until it is thrown, as othrwise
-    # the interpolation allocates and slows everything down a lot.
-    length(names_a) != length(names_b) && incompatible_dimension_error(names_a, names_b)
+function try_unify_names(names_a, names_b)
+    if names_a === names_b
+        return names_a
+    elseif length(names_a) != length(names_b)
```
oxinabox:
```suggestion
    elseif length(names_a) !== length(names_b)
```

src/name_core.jl
mcabbott: Good point, I've now moved these things around as suggested.

test/functions.jl
mcabbott: I didn't bother, as none of the other things in this file do, and there is a test for `names_are_unifiable` elsewhere.

src/name_core.jl
mcabbott: Can this matter? I guess there's no explicit restriction to `Tuple`s but surely these are both `Int`.

mcabbott:
Loads of allocation tests fail on nightly, btw, and a few unexpected passes. Can travis test on 1.4 yet?

src/name_core.jl
oxinabox: Yeah compiler likes `===` as it knows what it does.
So it can be better at optimizing it out of existence.

Idk if it happens in this case already.
But it's my habit for `nothing` and integers.

mcabbott: OK, I've changed all in this file.

PR #81, opened 2019-11-09
map + collect

### LINKED ISSUE #75: What should the behavour of map and collect be?
>  it is a more contentious feature ...

If you believe `map` & `collect`ed generators ought not to simply propagate names along, then perhaps open a discussion issue?

_Originally posted by @mcabbott in https://github.com/invenia/NamedDims.jl/pull/73#issuecomment-541540604_

mcabbott:
Closes #75.

Edit: test failure is an ambiguity on 1.0.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Requires = "0.5"
-Tracker = "0.2.2, 0.2.3, 0.2.4, 0.2.5"
+Tracker = "0.2.2"
 julia = "1"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.10"
+version = "0.2.11"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -11,7 +11,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Requires = "0.5"
-Tracker = "0.2.2"
+Tracker = "0.2.2, 0.2.3, 0.2.4, 0.2.5"
 julia = "1"
 
 [extras]
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -170,6 +170,8 @@ end
 Base.filter(f, A::NamedDimsArray{L,T,1}) where {L,T} = NamedDimsArray(filter(f, parent(A)), L)
 Base.filter(f, A::NamedDimsArray{L,T,N}) where {L,T,N} = filter(f, parent(A))
 
+
+# We overload collect on various kinds of `Generators` so that that can keep names.
 function Base.collect(x::Base.Generator{<:NamedDimsArray{L}}) where {L}
     data = collect(Base.Generator(x.f, parent(x.iter)))
     return NamedDimsArray(data, L)
@@ -177,7 +179,7 @@ end
 
 function Base.collect(x::Base.Generator{<:Iterators.Enumerate{<:NamedDimsArray{L}}}) where {L}
     data = collect(Base.Generator(x.f, enumerate(parent(x.iter.itr))))
-    NamedDimsArray(data, L)
+    return NamedDimsArray(data, L)
 end
 
 Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
@@ -187,5 +189,5 @@ Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsAr
 function collect_product(x)
     data = collect(Base.Generator(x.f, Iterators.product(unname.(x.iter.iterators)...)))
     all_names = tuple_cat(names.(x.iter.iterators)...)
-    NamedDimsArray(data, all_names)
+    return NamedDimsArray(data, all_names)
 end
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -142,7 +142,9 @@ end
 ################################################
 # map, collect
 
-# no explicit method needed for map(f, ::NamedDimsArray)
+# this method is only needed for Julia 1.0 in fact:
+Base.map(f, A::NamedDimsArray) = NamedDimsArray(map(f, parent(A)), names(A))
+
 for (T, S) in [
     (:NamedDimsArray, :AbstractArray),
     (:AbstractArray, :NamedDimsArray),
@@ -150,7 +152,8 @@ for (T, S) in [
     ]
     for fun in [:map, :map!]
 
-        @eval function Base.$fun(f, A::$T, B::$S, Cs::AbstractArray...)
+        # Here f::F where {F} is needed to avoid ambiguities in Julia 1.0
+        @eval function Base.$fun(f::F, A::$T, B::$S, Cs::AbstractArray...) where {F}
             data = $fun(f, unname(A), unname(B), unname.(Cs)...)
             new_names = unify_names(names(A), names(B), names.(Cs)...)
             return NamedDimsArray(data, new_names)
@@ -158,7 +161,7 @@ for (T, S) in [
 
     end
 
-    @eval function Base.foreach(f, A::$T, B::$S, Cs::AbstractArray...)
+    @eval function Base.foreach(f::F, A::$T, B::$S, Cs::AbstractArray...) where {F}
         data = foreach(f, unname(A), unname(B), unname.(Cs)...)
         unify_names(names(A), names(B), names.(Cs)...)
         return nothing
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -180,10 +180,6 @@ end
 
 function Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,Vararg{Any}}}})
     data = collect(Base.Generator(x.f, Iterators.product(unname.(x.iter.iterators)...)))
-    all_names = tuple_flatten(names.(x.iter.iterators)...)
+    all_names = tuple_cat(names.(x.iter.iterators)...)
     NamedDimsArray(data, all_names)
 end
-
-tuple_flatten(x::Tuple, ys::Tuple...) = (x..., tuple_flatten(ys...)...)
-tuple_flatten() = ()
-# @btime tuple_flatten((1, 2), (3, 4, 5), (6,)) # 0 allocations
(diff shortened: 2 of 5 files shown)


test/functions.jl
```diff
@@ -194,6 +194,62 @@ using Statistics
...
+        nda = NamedDimsArray([11 12; 21 22], (:x, :y))
+        ndv = NamedDimsArray([10, 20, 30], (:z,))
+
+        @test names([sqrt(x) for x in nda]) == (:x, :y)
+
+        @test names([x^i for (i,x) in enumerate(ndv)]) == (:z,)
+        @test names([x^i for (i,x) in enumerate(nda)]) == (:x, :y)
+
+        # Iterators.product -- has all names
+        @test names([x+y for x in nda, y in ndv]) == (:x, :y, :z)
```
oxinabox: woah, that is nice

oxinabox:
what about `map(f, ::Array, ::Array, ::NamedDimsArray)` ?
Do you think it is worth going beyond 2 ?

src/functions.jl
```diff
@@ -138,3 +138,49 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    @eval function Base.foreach(f, A::$T, B::$S, Cs::AbstractArray...)
+        data = foreach(f, unname(A), unname(B), unname.(Cs)...)
+        unify_names(names(A), names(B), names.(Cs)...)
+        return nothing
+    end
+end
+
+function Base.collect(x::Base.Generator{<:NamedDimsArray{L}}) where {L}
+    data = collect(Base.Generator(x.f, x.iter.data))
+    NamedDimsArray(data, L)
```
oxinabox:
```suggestion
    return NamedDimsArray(data, L)
```

src/functions.jl
```diff
@@ -138,3 +138,49 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+end
+
+function Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,Vararg{Any}}}})
+    data = collect(Base.Generator(x.f, Iterators.product(unname.(x.iter.iterators)...)))
+    all_names = tuple_flatten(names.(x.iter.iterators)...)
+    NamedDimsArray(data, all_names)
+end
+
+tuple_flatten(x::Tuple, ys::Tuple...) = (x..., tuple_flatten(ys...)...)
```
oxinabox: docstring would be good here.

src/name_core.jl
```diff
@@ -183,6 +183,10 @@ function unify_names(names_a, names_b)
     return compile_time_return_hack(ret)
 end
 
+unify_names(a) = a
+unify_names(a, b, cs...) = unify_names(unify_names(a,b), cs...)
+# @btime (()->unify_names((:a, :b), (:a, :_), (:_, :b)))()
```
oxinabox: add this to the allocation tests?

src/functions.jl
mcabbott: I start to think I should move this to name_core.jl, and rename it to something like `tuple_vcat`, if that sounds good.

mcabbott:
> what about `map(f, ::Array, ::Array, ::NamedDimsArray)` ?
> Do you think it is worth going beyond 2 ?

I guess going to 3 adds quite a bit of code, 10 cases instead of 3 if I'm thinking right, seems not worth it? (It does of course pick up names from the 3rd, so long as the 2nd has any names.)

oxinabox:
In TensorFlow.jl we did 3 for most things.
3 for me is the point at which is is obvious noone is going to do more.
but I personally rarely even remember the 2 arg version exists so that if probably fine.

Lets open an issue for it maybe so it is known that it doesn't work.
And maybe add a `@test_broken` ?
@iamed2  what do you think?

test/functions.jl
mcabbott: I made this work for 2 now, I mean if only the second factor is named.

src/functions.jl
```diff
@@ -138,3 +138,52 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+################################################
+# map, collect
+
+# no explicit method needed for map(f, ::NamedDimsArray)
+for (T, S) in [
+    (:NamedDimsArray, :AbstractArray),
+    (:AbstractArray, :NamedDimsArray),
+    (:NamedDimsArray, :NamedDimsArray),
+    ]
+    for fun in [:map, :map!]
```
iamed2: No need to use symbols for these; you can use the functions directly

mcabbott: > No need to use symbols for these; you can use the functions directly

Are you sure? 
```
julia> for fun in [sum, prod]
           @eval Base.$fun(x::String) = "yes"
       end
ERROR: syntax: invalid function name "Base.typeof(Base.sum)()"
Stacktrace:
 [1] top-level scope at REPL[14]:1
 [2] eval(::Module, ::Any) at ./boot.jl:331
 [3] top-level scope at REPL[14]:2

julia> sum("test")
ERROR: MethodError: no method matching +(::Char, ::Char)

julia> for fun in [:sum, :prod]
           @eval Base.$fun(x::String) = "yes"
       end

julia> sum("test")
"yes"
```

mcabbott:
Status is that it needs one more go at making tests pass on 1.0, BTW.

src/functions.jl
```diff
@@ -138,3 +138,52 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+
+        @eval function Base.$fun(f, A::$T, B::$S, Cs::AbstractArray...)
```
iamed2: For each function the generated functions are the same. You can create two functions `_map` and `_map!` and create the versions of `map` and `map!` that call the appropriate `_` function. That might be easier on the compiler if it decides it doesn't need to specialize on either of `T` or `S`.

src/functions.jl
iamed2: Nvm you're right, I'm thinking of the eval-less method which you can't do with type parameters in variables.

src/functions.jl
iamed2: I actually don't know if that would have an effect on load time. Up to you whether you want to do that.

mcabbott: You could split this, but will it help anything? I thought that what made this package fast was that the compiler specialised not just for every `T, S` but for every combination of names. (But am not certain how one would check this.)

Besides, it's a three-line function repeated 3 times, which doesn't seem like much.

src/functions.jl
mcabbott: OK, but `x::String` has no parameters?

mcabbott:
OK I think this is done, unless you'd like a 3rd-arg `@test_broken map(f, A, A, nda)` reminder in it.

test/functions.jl
```diff
@@ -204,21 +204,14 @@ using Statistics
         # this method only called based on first two arguments:
         @test names(map(+, parent(nda), parent(nda), nda)) == (:_, :_)
 
-        # one-arg forms work without adding anything, @which map(sqrt, nda) # Base
+        # one-arg forms work without adding anything... except on 1.0...
```
oxinabox: this line is not needed since it works now because it is in the package

src/functions.jl
```diff
@@ -142,23 +142,26 @@ end
...
 for (T, S) in [
     (:NamedDimsArray, :AbstractArray),
     (:AbstractArray, :NamedDimsArray),
     (:NamedDimsArray, :NamedDimsArray),
     ]
     for fun in [:map, :map!]
 
-        @eval function Base.$fun(f, A::$T, B::$S, Cs::AbstractArray...)
+        # Here f::F where {F} is needed to avoid ambiguities in Julia 1.0
+        @eval function Base.$fun(f::F, A::$T, B::$S, Cs::AbstractArray...) where {F}
```
oxinabox: It is confusing having single captial letter variables while also having single capital letter types.
Can we swap the `a::$T, b::$S, cs::AbstractArray...`

Project.toml
```diff
@@ -11,7 +11,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Requires = "0.5"
-Tracker = "0.2.2"
+Tracker = "0.2.2, 0.2.3, 0.2.4, 0.2.5"
```
oxinabox: This is not required.
`0.2.2` means "Anything greater than or equal to`0.2.2` and less than `0.3.0`

mcabbott: sorry, now fixed

oxinabox: no worries

src/functions.jl
```diff
@@ -138,3 +138,54 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    @eval function Base.foreach(f::F, a::$T, b::$S, cs::AbstractArray...) where {F}
+        data = foreach(f, unname(a), unname(b), unname.(cs)...)
+        unify_names(names(a), names(b), names.(cs)...)
+        return nothing
+    end
+end
+
+Base.filter(f, A::NamedDimsArray{L,T,1}) where {L,T} = NamedDimsArray(filter(f, parent(A)), L)
+Base.filter(f, A::NamedDimsArray{L,T,N}) where {L,T,N} = filter(f, parent(A))
+
```
oxinabox:
```suggestion


# We overload collect on various kinds of `Generators` so that that can keep names.
```

src/functions.jl
```diff
@@ -138,3 +138,54 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+function Base.collect(x::Base.Generator{<:NamedDimsArray{L}}) where {L}
+    data = collect(Base.Generator(x.f, parent(x.iter)))
+    return NamedDimsArray(data, L)
+end
+
+function Base.collect(x::Base.Generator{<:Iterators.Enumerate{<:NamedDimsArray{L}}}) where {L}
+    data = collect(Base.Generator(x.f, enumerate(parent(x.iter.itr))))
+    NamedDimsArray(data, L)
```
oxinabox:
```suggestion
    return NamedDimsArray(data, L)
```

src/functions.jl
```diff
@@ -138,3 +138,54 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+end
+
+Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
+Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:Any,<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
+Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
+
+function collect_product(x)
+    data = collect(Base.Generator(x.f, Iterators.product(unname.(x.iter.iterators)...)))
+    all_names = tuple_cat(names.(x.iter.iterators)...)
+    NamedDimsArray(data, all_names)
```
oxinabox:
```suggestion
    return NamedDimsArray(data, all_names)
```

PR #82, opened 2019-11-09
reshape simple case

mcabbott:
This is a tiny fix of `reshape` to do what I think you intended, and avoid this:
```
nda = NamedDimsArray(rand(2, 3), (:r, :c))
reshape(nda, 1, :) # 1×6 reshape(::NamedDimsArray{(:r, :c), …
```


nickrobinson251:
Nice catch!

PR #83, opened 2019-11-23
Switched out `names` for `dimnames`

### LINKED ISSUE #22: Used `Base.names`

Tokazama:
Fixes #22


nickrobinson251:
LGTM!

I am fine with `dimnames` as the name, but... before we go ahead I wanted to just raise the possibility of simply calling this `dims`, in case anyone else also preferred that? (But won't hold up this PR for more than a day)

oxinabox:
`dims` to me feels like it should return numbers

PR #84, opened 2019-11-23
Make keyword indexing fast

mcabbott:
This solves the `order_named_inds` problem by just making it generated, I fiddled around but couldn’t make other ways work. Not sure whether the body should then live in src/name_core.jl (as now) or move to src/wrapper_array.jl (where the function of this signature used to live).

Before:
```
julia> nda = NamedDimsArray(rand(2,2), (:x, :y));

julia> @btime $nda[1, 2];
  1.421 ns (0 allocations: 0 bytes)

julia> @btime $nda[x=1, y=2];
  40.455 ns (2 allocations: 64 bytes)
```
After:
```
julia> @btime $nda[x=1, y=2]
  1.420 ns (0 allocations: 0 bytes)
```

Edit: Note that this was the original motivation for #76. Which included a partial fix to get this down from 200ns.

### DIFF
--- a/src/NamedDims.jl
+++ b/src/NamedDims.jl
@@ -17,8 +17,8 @@ end
 # We use CoVector to workout if we are taking the tranpose of a tranpose etc
 const CoVector = Union{Adjoint{<:Any, <:AbstractVector}, Transpose{<:Any, <:AbstractVector}}
 
-include("wrapper_array.jl")
 include("name_core.jl")
+include("wrapper_array.jl")
 include("broadcasting.jl")
 include("functions.jl")
 include("functions_dims.jl")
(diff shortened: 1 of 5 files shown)


mcabbott:
This needs re-checking BTW, I may have broken it with force push, or perhaps just got confused locally. But assuming it works, thoughts on the approach?

oxinabox:
I like it.

Maybe we could make it that a `Val{L}` where L is a triple.
Rather that taking the `NamedDimArray` directly.

I have some aspirations of keeping the name_core.jl file reasonably independent of the use in Arrays..to one day use on idk multivariate distributions or something.

But other than that I think we should do this.

mcabbott:
Excellent. I made it use `Val` as suggested, and thus could revert the organisation of what's in what file. On Julia 1.3, and on 1.03, it's now fast as claimed above.

PR #87, opened 2019-12-19
Requires is now 1.0

rofinn:


Project.toml
```diff
@@ -10,7 +10,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-Requires = "0.5"
```
oxinabox: Can we allow both?

rofinn: I guess not. AFAICT, there's just some performance benefits in 1.0.

rofinn:
Updated both... feel free to squash.

oxinabox:
LGTM

nickrobinson251:
Thanks!

Issue #89, opened 2020-01-22
Allocation tests post-1.3


mcabbott:
Many of the allocation tests fail on master (and on 1.4), essentially because this is nonzero:
```
julia> @allocated (() -> (1+1))()
62992

julia> VERSION
v"1.5.0-DEV.88"
```
On slack, kristofferc points me to https://github.com/JuliaLang/julia/pull/33717, after which the `@allocated` macro no longer excludes compilation cost. So we will need to adjust how we test.

iamed2:
I just came here to make this issue, based on your own Slack conversation, because I'm daft 😆

iamed2:
This is true:
```
julia> f = () -> (1+1)
#17 (generic function with 1 method)

julia> @allocated f()
61808

julia> @allocated f()
0
```

So I think all we'd need is a wrapper macro that would run code that returns a function, call it without arguments, then call `@allocated` on it, since AFAICT our only problematic case just defines an anonymous function and calls it without arguments.

mcabbott:
One could also adapt `@btime` to make a `@ballocated`. 

Or perhaps just dispense with these `(() -> ...)()` things, since the functions inside should already have been called before? Does zero allocations imply that constant prop. is working correctly?

iamed2:
Yes I believe those anonymous functions are written that way in order to test that constant prop. is working.

PR #91, opened 2020-01-24
Fix #90

### LINKED ISSUE #90: Should this allocate?
```julia
using NamedDims, BenchmarkTools
X = randn(500, 30);
Y = NamedDimsArray(X, (:N, :D));
@benchmark NamedDimsArray($Y, (:N, :D))
BenchmarkTools.Trial:
  memory estimate:  256 bytes
  allocs estimate:  5
  --------------
  minimum time:     538.402 ns (0.00% GC)
  median time:      548.471 ns (0.00% GC)
  mean time:        585.369 ns (2.63% GC)
  maximum time:     17.383 μs (95.24% GC)
  --------------
  samples:          10000
  evals/sample:     189
```

mcabbott:
Before: 
```julia
julia> X = randn(500, 30);

julia> Y = NamedDimsArray(X, (:N, :D));

julia> @btime NamedDimsArray($X, (:N, :D)); # first construction also slow
  515.563 ns (6 allocations: 272 bytes)

julia> @btime NamedDimsArray{(:N, :D)}($X);
  5.691 ns (1 allocation: 16 bytes)

julia> @btime NamedDimsArray($Y, (:N, :D)); # as in issue 90
  511.484 ns (5 allocations: 256 bytes)

julia> @btime NamedDimsArray{(:N, :D)}($Y);
  1.421 ns (0 allocations: 0 bytes)

julia> const cnda = NamedDimsArray([10 20; 30 40], (:x, :y)); # from allocation tests

julia> 0 == @allocated NamedDimsArray(cnda, (:x, :y))
true

julia> NamedDimsArray(Y, (:N, :D));

julia> @allocated NamedDimsArray(Y, (:N, :D)) # without const
256

julia> @allocated (() -> NamedDimsArray(Y, (:N, :D)))()
256
```
Afterwards, these are all fast. 

The allocation test [from the tests](https://github.com/invenia/NamedDims.jl/blob/master/test/wrapper_array.jl#L178) is still happy, and the one without `const` is the same. 

But `@btime` we can trust, right? Then it closes #90.

### DIFF
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -44,10 +44,8 @@ function NamedDimsArray{L}(orig::AbstractArray{T, N}) where {L, T, N}
     return NamedDimsArray{L, T, N, typeof(orig)}(orig)
 end
 
-function NamedDimsArray(orig::AbstractArray{T, N}, names::NTuple{N, Symbol}) where {T, N}
-    return NamedDimsArray{names}(orig)
-end
-NamedDimsArray(orig::AbstractVector, name::Symbol) = NamedDimsArray(orig, (name,))
+@inline NamedDimsArray(orig::AbstractArray, names::Tuple) = NamedDimsArray{names}(orig)
+@inline NamedDimsArray(orig::AbstractVector, name::Symbol) = NamedDimsArray(orig, (name,))
 
 # Name-asserting constructor (like renaming, but only for wildcards (`:_`).)
 NamedDimsArray{L}(orig::NamedDimsArray{L}) where L = orig


oxinabox:
Looks fine to me.


We must need to add a test somewhere.
Since tests were passing when #90 was openned

mcabbott:
Perhaps testing with `@allocated` isn't actually good enough, since it fails to spot this? Should they should all become `@btime` or something?

oxinabox:
Any idea what we should do to test this?

mcabbott:
I don't really understand why `@allocated` disagrees with `@btime`. But if we trust the latter, then I think it would not be so hard to hack together an `@belapsed` by looking at `@btime`. Perhaps you could even make it run just a few iterations so that tests don't take ages. Probably that is where #89 should go.

oxinabox:
Hmm, ok. I think should merge this and deal with tests in #89

PR #93, opened 2020-02-08
Support FFTs

mcabbott:
Since @oxinabox was complaining of having nothing to review... this may need a little updating and a little polish. 

The idea was that FFT ought to be written like `fft(A, :time => :freq)` since obviously those aren't the same thing, and you would never want to end taking a dot product of them etc. And if you don't specify a name, then `:x∿` is a character that behaves well, I think.


src/functions_math.jl
```diff
@@ -108,3 +113,131 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
+            $($fun)(A, :time => :freq, :x => :kx)
+
+        Acting on a `NamedDimsArray`, this specifies to take the transform along the dimensions
+        named `:time, :x`, and return an array with names `:freq` and `:kx` in their places.
+
+        If new names are not given, then the default is `:x => :x∿` and `:x∿ => :x`,
+        applied to all dimensions, or to those specified as usual, e.g. `$($fun)(A, (1,2))`
+        or `$($fun)(A, :time)`. The symbol "∿" can be typed by `\\sinewave<tab>`.
+        """
+        function AbstractFFTs.$fun(A::NamedDimsArray{L,T,N}, p::Pair{Symbol,Symbol}, ps::Pair{Symbol,Symbol}...) where {L,T,N}
```
oxinabox: Thoughs on the pro and con of using kwargs to rename?
```
fft((A;  x=:xf, y=:yf)
```

src/functions_math.jl
```diff
@@ -108,3 +113,131 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
+        `:name∿` in the transformed dimension.
+        """
+        function AbstractFFTs.$plan_fun(A::NamedDimsArray, dims = ntuple(d->d, ndims(A)); kw...)
+            numerical_dims = dim(A, dims)
+            return AbstractFFTs.$plan_fun(parent(A), numerical_dims; kw...)
+        end
+
+        AbstractFFTs.$plan_fun(A::NamedDimsArray, p::Pair{Symbol,Symbol}, ps::Pair{Symbol,Symbol}...) =
+            throw(ArgumentError(string($plan_fun, "(::NamedDimsArray, ::Pair...) is not implemented. ",
+                "FFT plans can only specify an input name, not an output.")))
```
oxinabox: Per BlueStyle https://github.com/invenia/BlueStyle/blob/master/.mailmap#method-definitions

> Only use short-form function definitions when they fit on a single line

I suspect this code was written before we adopted the style-guide ?

src/functions_math.jl
```diff
@@ -108,3 +113,131 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
+        return NamedDimsArray(data, newL)
+    end
+
+end
+
+if VERSION > v"1.1"
+    _hasproperty(x, s)::Bool = hasproperty(x, s)
+else
+    _hasproperty(x, s)::Bool = Base.sym_in(s, propertynames(x))
+end
```
oxinabox: this should move to `src/compat.jl`
Or possibly we can just make use of the **Compat.jl** package for it?

src/name_core.jl
```diff
@@ -318,11 +318,32 @@ end
...
+    end
+    return compile_time_return_hack(out)
+end
+# @btime NamedDims.replace_names((:a, :b), :b => :c) # 1.420 ns (0 allocations: 0 bytes)
+
+function replace_names(dimnames::Tuple, pair::Pair, pairs::Pair...)
+    step_one = replace_names(dimnames, pair)
+    return replace_names(step_one, pairs...)
+end
+# @btime NamedDims.replace_names((:a, :b), :b => :c, :a => :z) # 1.420 ns (0 allocations: 0 bytes)
```
oxinabox: Can we move these into the tests?

oxinabox:
This would be the first time this package actually decided on a name for a thing on its own.
Its never done that in any other PR before.
Only ever inserted the wildcard name.

I think the naming done in this PR make sense though.

My only worry is if we could be creating name conflicts (admittedly the names would have had to be pretty weird to end up in that situation).

---
unrelated:

It seems like this name actually works fine as a kwarg which is good.
```
julia> foo(;kwargs...) = kwargs
foo (generic function with 1 method)

julia> foo(x∿=4)
pairs(::NamedTuple) with 1 entry:
  :x∿ => 4
```

src/functions_math.jl
mcabbott: I can just always use ` Base.sym_in` inline. (I moved it out to see if it mattered for speed.)

The treatment of `plan_fft` seems a bit awkward to me. The alternative it to return a new struct `NamedPlan` or something, which would explicitly encode the names & dimensions, and make its application zero-cost again. But perhaps this adds more code than it deserves, thoughts?

src/functions_math.jl
mcabbott: Or was doodled up for my own amusement. Easy to fix.

src/name_core.jl
mcabbott: Sure

src/functions_math.jl
mcabbott: I thought about that, but concluded that it was too confusing about which one is the old & the new name.

src/functions_math.jl
oxinabox: > I can just always use Base.sym_in inline. (I moved it out to see if it mattered for speed.)

That would be worse for readability.
just a 
```
if isdefined(Base, :hasproperty)
    hasproperty(x, s)::Bool = Base.sym_in(s, propertynames(x))
end
```

in the compat file.
(or check Compat.jl)

mcabbott:
Yes I was quite pleased to find `∿`, it's relevant, you can type it, and it's an ordinary letter as far as Julia's concerned. I can't find a single example of anyone using it in code, so I doubt there will be collisions.

src/functions_math.jl
oxinabox: I think `NamedPlan` make sense, but i don't know FFT well

mcabbott: OK will tidy up the struct version then.

mcabbott: Will get to this eventually, BTW. The struct version had a few unexpected complications.

mcabbott:
BTW, the status here is that the version with `struct NamedPlan` kept getting more complicated, and I wasn't exactly sure what its rules should be. (I never pushed this, it seems.)

The simpler version here with `Base.:*(plan::AbstractFFTs.$plan_type, A::NamedDimsArray{L,T,N})` (in which the plan can't explicitly carry names) can probably be improved using  https://github.com/JuliaMath/FFTW.jl/pull/119, which makes the dims on which it acts part of the type.

src/functions_math.jl
mcabbott: OK, I trashed the whole `NamedPlan` business, it got complicated. We can re-invent it if someone uses this & needs it, maybe make the basic case work first.

mcabbott:
Maybe this is done. 

It has only iffy support for `plan_fft`, and none at all for `mul!(out, plan, input)` or `fft!` or `plan_fft!`. I guess `fft!` would not be hard to add, but the others... can be done later if someone wants them.

src/name_core.jl
mcabbott: This is now deleted, in favour of `_rename` from another PR.

src/functions_math.jl
```diff
@@ -127,3 +132,152 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
 end
+
+################################################
+# FFT
+
+for fun in (:fft, :ifft, :bfft, :rfft, :irfft, :brfft)
+    plan_fun = Symbol(:plan_, fun)
+
+    if fun in (:irfft, :brfft) # These take one more argument, a size
+        extra = (:(d::Integer),)
```
oxinabox, line 143: Should this be  called `extra_args` or something?

src/functions_math.jl
```diff
@@ -127,3 +132,152 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
+@btime plan_fft($nda, (:k,:l)) # 18.393 μs      #    42.622 μs
+
+@btime $pp * $nda             #   9.029 μs      # 955.534 μs
+@btime $pp * $(parent(nda))   #     125.652 ns  # 943.517 μs
+@code_warntype pp * nda       # pretty awful!
+
+ndc = NamedDimsArray(zeros(4,4) .+ im, (:p, :q))
+mul!(ndc.data, pp, nda.data .+ 0im)
+mul!(ndc, pp, nda .+ 0im) # no method
+=#
```
oxinabox, line 253: This commented out section probable needs to be deleted?

src/functions_math.jl
```diff
@@ -127,3 +132,152 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
     end
     return compile_time_return_hack(names)
 end
+
+################################################
+# FFT
```
oxinabox, line 137: Maybe we should move this all to its own file?
and then the test to match?

src/functions_math.jl
```diff
@@ -127,3 +132,152 @@ function symmetric_names(L::Tuple{Symbol,Symbol}, dims::Integer)
...
+
+for fun in (:fft, :ifft, :bfft, :rfft, :irfft, :brfft)
+    plan_fun = Symbol(:plan_, fun)
+
+    if fun in (:irfft, :brfft) # These take one more argument, a size
```
oxinabox, line 142:
```suggestion
    if fun in (:irfft, :brfft)  # These take one more argument, a size
```

Two spaces before end of line comments, per BlueStyle

oxinabox (approved):
Ok I am once again getting toward the bottom of myh to review stack (though i think there are still a dozen things below this)

This LGTM.
I have granted you merge rights on the repo.
so you can merge this when ready.
All the usual [ColPrac](https://github.com/SciML/ColPrac) things apply about not merging your own PR without review (give ColPrac a read if you haven't already)

src/functions_math.jl
mcabbott: Will do.

src/functions_math.jl
mcabbott: I changed it, to also include a string to insert into docstrings.

src/functions_math.jl
mcabbott: Done.

mcabbott:
OK, thanks! 

If I understand that right I'm supposed to merge this, although perhaps I wait until solid CI. Will JuliaRegistrator also take orders from me now?

nickrobinson251:
> I'm supposed to merge this

Yeah, merge your own PRs once they're approved and CI passes, then register the new version (fingers crossed Registrator will work for you now) :) 

thanks!

oxinabox:
Urg, i bet Registrator won't work.

mcabbott:
Done! 

#149 has a version bump & is approved, I will merge it too & then this can be tagged.

PR #94, opened 2020-02-13
Replace name asserting constructor with refine names

oxinabox:
This has never sat right with me.
That calling `NamedDimsArray` on a `NamedDimsArray` was the way to coerce wildcard names to have a particular name."

[PyTorch calls this operation `refine_names`](https://pytorch.org/docs/stable/named_tensor.html#torch.Tensor.refine_names)
i think this is a much better name for the function

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.15"
+version = "0.2.16"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ For `nda = NamedDimsArray{(:x, :y, :z)}(rand(10, 20, 30))`.
 
  - Indexing: `nda[y=2]` is the same as `nda[x=:, y=2, z=:]` which is the same as `nda[:, 2, :]`.
  - Functions taking a `dims` keyword: `sum(nda; dims=:y)` is the same as `sum(nda; dims=2)`.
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -18,10 +18,11 @@ For `nda = NamedDimsArray{(:x, :y, :z)}(rand(10, 20, 30))`.
 
  - Indexing: `nda[y=2]` is the same as `nda[x=:, y=2, z=:]` which is the same as `nda[:, 2, :]`.
  - Functions taking a `dims` keyword: `sum(nda; dims=:y)` is the same as `sum(nda; dims=2)`.
- - Renaming: `rename(nda, new_names)` returns a new `NamedDimsArray` with the `new_names` but still wrapping the same data.
+ - Renaming: `rename(nda, new_names)` returns a new `NamedDimsArray` with the `new_names` but still wrapping the same data
```
nickrobinson251: Where did the full stop go?

oxinabox: Weird

oxinabox:
```suggestion
 - Renaming: `rename(nda, new_names)` returns a new `NamedDimsArray` with the `new_names` but still wrapping the same data.
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.15"
+version = "0.2.16"
```
omus: Seems like a feature bump

oxinabox: It is a nonbreaking change.
Pre-1.0, the minor version number is only bumped for breaking changes.
(Haven't we had this discussion?)

omus: Looking over semver again there's nothing even stating that a major version zero patch release has to be non-breaking.

omus: I tend to stick to patch releases are fixes while minor releases are features

oxinabox: Lets take this to slack.

PR #99, opened 2020-02-21
dimnames(A, d)

mcabbott:
If `dimnames` is the official way to extract names, then perhaps `dimnames(A, d)` ought to work like `size(A, d)`.

### DIFF
--- a/src/name_operations.jl
+++ b/src/name_operations.jl
@@ -39,8 +39,13 @@ unname(x::AbstractArray) = x
 
 """
     dimnames(A) -> Tuple
+    dimnames(A, d) -> Symbol
 
-Return the names of all the dimensions of the array `A`.
+Return the names of all the dimensions of the array `A`, 
+or just the one for the `d`-th dimension. 
+
+Gives wildcards `:_` if this is not a `NamedDimsArray`.
+And like `size(A, d)`, it allows `d > ndims(A)`.
 """
 dimnames(::Type{<:NamedDimsArray{L}}) where {L} = L
 dimnames(::Type{<:AbstractArray{T, N}}) where {T, N} = ntuple(_->:_, N)
--- a/src/name_operations.jl
+++ b/src/name_operations.jl
@@ -57,7 +57,7 @@ function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
     elseif d>N
         return :_
     else
-        error("dimnames: dimension out of range")
+        throw(DimensionMismatch("dimnames: dimension out of range"))
     end
 end
 dimnames(x::T, d::Integer) where T<:AbstractArray = dimnames(T, d)
--- a/src/name_operations.jl
+++ b/src/name_operations.jl
@@ -42,6 +42,17 @@ unname(x::AbstractArray) = x
 
 Return the names of all the dimensions of the array `A`.
 """
-dimnames(::Type{<:NamedDimsArray{L}}) where L = L
+dimnames(::Type{<:NamedDimsArray{L}}) where {L} = L
 dimnames(::Type{<:AbstractArray{T, N}}) where {T, N} = ntuple(_->:_, N)
 dimnames(x::T) where T<:AbstractArray = dimnames(T)
+
+function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
+    if d in 1:N
+        return dimnames(AT)[d]
+    elseif d>N
+        return :_
+    else
+        error("dimnames: dimension out of range")
+    end
+end
+dimnames(x::T, d::Integer) where T<:AbstractArray = dimnames(T, d)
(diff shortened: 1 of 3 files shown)


src/name_operations.jl
```diff
@@ -39,9 +39,25 @@ unname(x::AbstractArray) = x
...
 dimnames(::Type{<:AbstractArray{T, N}}) where {T, N} = ntuple(_->:_, N)
 dimnames(x::T) where T<:AbstractArray = dimnames(T)
+
+function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
+    if d in 1:N
+        return dimnames(AT)[d]
+    elseif d>N
+        return :_
+    else
+        error("dimnames: dimension out of range")
```
nickrobinson251: I think we'd want this to be a BoundsError, like
```julia
julia> size(1:3, 0)
ERROR: BoundsError: attempt to access (3,)
  at index [0]
```

I think we can have that with 
```julia
function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
    return d > N ? :_ : dimnames(AT)[d]
end
```
```julia
julia> dimnames([10 20; 30 40], 0)
ERROR: BoundsError: attempt to access (:_, :_)
  at index [0]
```

mcabbott: I followed this behaviour:
```
julia> try size(ones(2),0) catch e e end
ErrorException("arraysize: dimension out of range")
```
but indeed it seems the Base isn't consistent about this.

nickrobinson251: Fair enough. Silly Base 😄 I think plain `ErrorExceptions` are less helpful, and the BoundsError seems right here, so i'd prefer we go with that, if you're fine with it?

Otherwise I think this looks good 👍

nickrobinson251:
remember to bump the version number too :)

src/name_operations.jl
```diff
@@ -39,9 +39,25 @@ unname(x::AbstractArray) = x
...
-dimnames(::Type{<:NamedDimsArray{L}}) where L = L
+dimnames(::Type{<:NamedDimsArray{L}}) where {L} = L
 dimnames(::Type{<:AbstractArray{T, N}}) where {T, N} = ntuple(_->:_, N)
 dimnames(x::T) where T<:AbstractArray = dimnames(T)
+
+function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
+    if d in 1:N
+        return dimnames(AT)[d]
+    elseif d>N
+        return :_
```
oxinabox: What do we want this case?

mcabbott: `axes(1:2, 99)` and `size(1:2, 99)` aren't errors, so I thought this should likewise give something. Is there another logical choice? 

Also, if `v = NamedDimsArray(1:2, :a)` the it's nice if `dimnames(v, 2) == dimnames(permutedims(v),1)`.

oxinabox: Fair points, lets put that into a comment in the source

src/name_operations.jl
```diff
@@ -39,9 +39,25 @@ unname(x::AbstractArray) = x
...
+Gives wildcards `:_` if this is not a `NamedDimsArray`.
+And like `size(A, d)`, it allows `d > ndims(A)`.
 """
-dimnames(::Type{<:NamedDimsArray{L}}) where L = L
+dimnames(::Type{<:NamedDimsArray{L}}) where {L} = L
 dimnames(::Type{<:AbstractArray{T, N}}) where {T, N} = ntuple(_->:_, N)
 dimnames(x::T) where T<:AbstractArray = dimnames(T)
+
+function dimnames(AT::Type{<:AbstractArray{T,N}}, d::Integer) where {T,N}
+    if d in 1:N
```
oxinabox: I think this is easier to read:
```suggestion
    if 1 <= d <= N
```

src/name_operations.jl
```diff
@@ -39,9 +39,25 @@ unname(x::AbstractArray) = x
...
+        return dimnames(AT)[d]
+    elseif d>N
```
oxinabox: Spacing per BlueStyle

```suggestion
    elseif d > N
```

src/name_operations.jl
mcabbott: I added it to the tests, so that it can't break. Hope that's good?

src/name_operations.jl
```diff
@@ -39,9 +39,25 @@ unname(x::AbstractArray) = x
...
 """
     dimnames(A) -> Tuple
+    dimnames(A, d) -> Symbol
 
-Return the names of all the dimensions of the array `A`.
+Return the names of all the dimensions of the array `A`, 
+or just the one for the `d`-th dimension. 
+
+Gives wildcards `:_` if this is not a `NamedDimsArray`.
+And like `size(A, d)`, it allows `d > ndims(A)`.
```
oxinabox:
```suggestion
Like `size(A, d)`, it allows `d > ndims(A)`, in this case all the trailing dimension are given the wildcard name (`:_`).
```

oxinabox:
thanks

PR #103, opened 2020-03-16
Revert `refine_names`

nickrobinson251:
- This reverts https://github.com/invenia/NamedDims.jl/pull/94
- The motivation for #94 was to make `NamedDimsArray(::NamedDimsArray, names)` return a NDA wrapped in another NDA, i.e. like `PermutedDimsArray` (we didn't switch to that functionality yet, the idea was just "that would be the expected behaviour"):
```julia
julia> p = PermutedDimsArray(rand(2, 3), (1, 2));

julia> q = PermutedDimsArray(p, (1, 2))
2×3 PermutedDimsArray(PermutedDimsArray(::Array{Float64,2}, (1, 2)), (1, 2)) with eltype Float64:
 0.899565  0.0589341  0.350773
 0.430388  0.750283   0.694883
````
- Or rather, the motivation was that `NamedDimsArray(::NamedDimsArray, names)` _not_ doing that and instead returning a non-nested NDA (with the updated "names" in place of any wild-cards) was "surprising behaviour"
- But introducing `refine_names` as the way to "refine names", had an unfortunate side effect: `NamedDimsArray` was no longer the safe way to construct a NDA from a `<:AbstractArray`, for that we now had to use `refine_names`. This made `refine_names` the primary constructor for NDAs. 
- Having seen `refine_names` become the primary constructor in a number of (private) package, where functions take a `<:AbstractArray` that may or may not be a NDA, it is much more opaque what is going on. And was a step backwards in terms of readability of the code.
- This PR say: it is better to have `NamedDimsArray` be the primary interface, and the behaviour of `NamedDimsArray(::NamedDimsArray, names)` is not particularly surprising (nesting NDAs is wierd, why are there two layers of names?! 😕).

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -60,15 +60,13 @@ whether they are using `NamedDimsArray`s or not.
 While also being able to use `NamedDimsArray`s internally in its definition;
 and also getting the assertion when a `NamedDimsArray` _is_  passed in, that it has the
 expected dimensions.
-The way to do this is to use the `refine_names(x, expected_names)`.
-This will: apply the names to a unnamed array, or to unnamed dimensions in a `NamedDimsArray,
-while also asserting any names that were given are correct.
-
+The way to do this is to call the `NamedDimsArray` constructor, with the expected names
+within the function.
```
oxinabox:
```suggestion
within the function.
This operation corresponds to [PyTorch's `refine_names`](https://pytorch.org/docs/stable/named_tensor.html#torch.Tensor.refine_names).
```

nickrobinson251:
@mcabbott this make sense to you?

mcabbott:
Yes, I prefer this behaviour. 

I also think that `PermutedDimsArray(PermutedDimsArray(A,...` should simplify to one wrapper. Why would two ever be desired? And one still ensures the type returned matches the function's name.

src/NamedDims.jl
```diff
@@ -7,7 +7,7 @@ using Pkg
 using Requires
 using Statistics
 
-export NamedDimsArray, dim, refine_names, rename, unname, dimnames
+export NamedDimsArray, dim, rename, unname, dimnames
```
mcabbott: Perhaps it should still be exported though, else you won't get the `@deprecate`d function?

nickrobinson251: yeah, that's right. I believe it is exported still via the `true` arg in `@deprecate refine_names NamedDimsArray true`

mcabbott: Oh I didn't know that, sorry. All good then!

nickrobinson251: yeah, no worries. magical bools are magical. and it made me go double-check which was good :)

PR #104, opened 2020-03-17
Indexing by BitArray{3}, etc.

### LINKED ISSUE #70: Indexing by BitArray{N}
This doesn’t error, but why should it have name `:x` not `:z`? Probably it should have neither (ref #69):
```
julia> nda = NamedDimsArray(rand(1:9, 3,3,3), (:x, :y, :z));

julia> nda[rand(3,3,3) .> 0.5]
17-element NamedDimsArray{(:x,),Int64,1,Array{Int64,1}}:
```
The case of `BitArray{1}` works correctly:
```
julia> nda[rand(3) .> 0.3, 1, :]
2×3 NamedDimsArray{(:x, :z),Int64,2,Array{Int64,2}}:
```

mcabbott:
Closes #70

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -306,7 +306,7 @@ end
 
 function remaining_dimnames_from_indexing(
     dimnames::Tuple{<:Any, <:Any, Vararg}, inds::Tuple{T}
-) where T <: Union{BitArray, AbstractVector{<:CartesianIndex}}
+) where T <: Union{AbstractArray{Bool}, AbstractVector{<:CartesianIndex}}
     return nothing
 end
 
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -268,12 +268,18 @@ is_noninteger_type(::Any) = true
 
 """
     remaining_dimnames_from_indexing(dimnames::Tuple, inds)
-Given a tuple of dimension names
-and a set of index expressesion e.g `1, :, 1:3, [true, false]`,
-determine which are not dropped.
-Dimensions indexed with scalars are dropped
+
+Given a tuple of dimension names,
+and a tuple of indices e.g `(1, :, 1:3, [true, false])`,
+this drops those indexed with scalars or `CartesianIndex`,
+inserts `:_` for `newaxis = [CartesianIndex{0}()]`,
+and returns another tuple of names.
+
+It may also return `nothing`, to indicate that names should be dropped.
+This happens e.g. when indexing a matrix by a `BitArray{2}` such as `mat[mat .> 0.5]`,
+or by a vector of `CartesianIndex`es.
 """
-@generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds)
+@generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds::Tuple)
     # 0-Allocation see:
     # `@btime (()->remaining_dimnames_from_indexing((:a, :b, :c), (:,390,:)))()``
     keep_names = []
@@ -291,6 +297,16 @@ Dimensions indexed with scalars are dropped
         end
     end
     return Expr(:call, :compile_time_return_hack, Expr(:tuple, keep_names...))
+end
+
+for T in [
+    :BitArray,
+    # :(NamedDimsArray{L,T,N,<:BitArray} where {L,T,N}), # defined in wrapper_array.jl
+    :(AbstractVector{<:CartesianIndex}),
+]
+    # These only apply to ndims(A) >= 2
+    @eval remaining_dimnames_from_indexing(dimnames::Tuple{Symbol, Symbol, Vararg{Symbol}},
+    inds::Tuple{$T}) = nothing
 end
 
 """
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -275,7 +275,7 @@ this drops those indexed with scalars or `CartesianIndex`,
 inserts `:_` for `newaxis = [CartesianIndex{0}()]`,
 and returns another tuple of names.
 
-It may also return `nothing`, to indicate that names should be dropped.
+It will return an empty tuple to indicate that all names should be dropped.
 This happend for scalar indexing by integers, or one `CartesianIndex`.
 It also happens e.g. when indexing a matrix by a `BitArray{2}` such as `mat[mat .> 0.5]`:
 this returns a vector, the same as vec(mat)[vec(mat .> 0.5)], whose dimension isn't any
@@ -306,13 +306,13 @@ of the original dimensions, hence has no name.
     return Expr(:call, :compile_time_return_hack, Expr(:tuple, keep_names...))
 end
 
-remaining_dimnames_from_indexing(dn::Tuple, inds::Tuple{Vararg{<:Integer}}) = nothing
-remaining_dimnames_from_indexing(dn::Tuple, ci::Tuple{CartesianIndex}) = nothing
+remaining_dimnames_from_indexing(dn::Tuple, inds::Tuple{Vararg{<:Integer}}) = ()
+remaining_dimnames_from_indexing(dn::Tuple, ci::Tuple{CartesianIndex}) = ()
 
 function remaining_dimnames_from_indexing(
     dimnames::Tuple{<:Any, <:Any, Vararg}, inds::Tuple{T}
 ) where T <: Union{Base.LogicalIndex, AbstractVector{<:CartesianIndex}}
-    return nothing
+    return ()
 end
 
 """
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -282,12 +282,6 @@ or by a vector of `CartesianIndex`es.
 @generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds::Tuple)
     # 0-Allocation see:
     # `@btime (()->remaining_dimnames_from_indexing((:a, :b, :c), (:,390,:)))()``
-    if length(inds.parameters) == 1 && length(dimnames.parameters) >= 2
-        type = inds.parameters[1]
-        if type <: BitArray || type <: AbstractVector{<:CartesianIndex}
-            return nothing
-        end
-    end
     keep_names = []
     dim_num = 0
     for type in inds.parameters
@@ -308,6 +302,12 @@ or by a vector of `CartesianIndex`es.
         end
     end
     return Expr(:call, :compile_time_return_hack, Expr(:tuple, keep_names...))
+end
+
+function remaining_dimnames_from_indexing(
+    dimnames::Tuple{<:Any, <:Any, Vararg}, inds::Tuple{T}
+) where T <: Union{BitArray, AbstractVector{<:CartesianIndex}}
+    return nothing
 end
 
 """
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -282,6 +282,12 @@ or by a vector of `CartesianIndex`es.
 @generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds::Tuple)
     # 0-Allocation see:
     # `@btime (()->remaining_dimnames_from_indexing((:a, :b, :c), (:,390,:)))()``
+    if length(inds.parameters) == 1 && length(dimnames.parameters) >= 2
+        type = inds.parameters[1]
+        if type <: BitArray || type <: AbstractVector{<:CartesianIndex}
+            return nothing
+        end
+    end
     keep_names = []
     dim_num = 0
     for type in inds.parameters
@@ -289,24 +295,19 @@ or by a vector of `CartesianIndex`es.
             dim_num += 1
         elseif type <: CartesianIndex
             dim_num += type.parameters[1]
-        elseif type == Array{CartesianIndex{0},1}
+        elseif type <: AbstractVector{CartesianIndex{0}}
             push!(keep_names, QuoteNode(:_))
+        elseif type <: AbstractArray{<:Integer} && ndims(type) > 1
+            dim_num += 1
+            for _ in 1:ndims(type)
+                push!(keep_names, QuoteNode(:_))
+            end
         else
             dim_num += 1
             push!(keep_names, :(getfield(dimnames, $dim_num)) )
         end
     end
     return Expr(:call, :compile_time_return_hack, Expr(:tuple, keep_names...))
-end
-
-for T in [
-    :BitArray,
-    # :(NamedDimsArray{L,T,N,<:BitArray} where {L,T,N}), # defined in wrapper_array.jl
-    :(AbstractVector{<:CartesianIndex}),
-]
-    # These only apply to ndims(A) >= 2
-    @eval remaining_dimnames_from_indexing(dimnames::Tuple{Symbol, Symbol, Vararg{Symbol}},
-    inds::Tuple{$T}) = nothing
 end
 
 """
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -276,8 +276,10 @@ inserts `:_` for `newaxis = [CartesianIndex{0}()]`,
 and returns another tuple of names.
 
 It may also return `nothing`, to indicate that names should be dropped.
-This happens e.g. when indexing a matrix by a `BitArray{2}` such as `mat[mat .> 0.5]`,
-or by a vector of `CartesianIndex`es.
+This happend for scalar indexing by integers, or one `CartesianIndex`.
+It also happens e.g. when indexing a matrix by a `BitArray{2}` such as `mat[mat .> 0.5]`:
+this returns a vector, the same as vec(mat)[vec(mat .> 0.5)], whose dimension isn't any
+of the original dimensions, hence has no name.
 """
 @generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds::Tuple)
     # 0-Allocation see:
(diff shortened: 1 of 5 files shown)


test/wrapper_array.jl
```diff
@@ -72,6 +72,24 @@ end
         @test size(nda[CartesianIndex(1, 1), newaxis]) == (1, )
         @test nda[CartesianIndex(1, 1), newaxis][1] == nda[1, 1]
     end
+
+    # https://github.com/invenia/NamedDims.jl/issues/70
+    @testset "BitArray etc" begin
```
oxinabox: Can we add more subtestset sets?

test/wrapper_array.jl
```diff
@@ -72,6 +72,24 @@ end
...
+        nda = NamedDimsArray(rand(1:9, 3,3,3), (:x, :y, :z))
+        @test dimnames(nda[rand(3) .> 0.3, 1, :]) == (:x, :z)
+        @test dimnames(nda[[1,3], 1, :]) == (:x, :z)
+
+        # flattening operations should drop names
+        @test dimnames(nda[rand(3,3,3) .> 0.3]) == (:_,)
+        @test dimnames(nda[nda .> 3]) == (:_,)
+        @test dimnames(nda[findall(iseven, nda)]) == (:_,)
```
oxinabox: This one seems different to others?

src/wrapper_array.jl
```diff
@@ -168,6 +168,15 @@ for f in (:getindex, :view, :dotview)
         @propagate_inbounds function Base.$f(a::NamedDimsArray, inds...)
             # Some nonscalar case, will return an array, so need to give that names.
             data = Base.$f(parent(a), inds...)
+            if ndims(a) > 1 && inds isa Union{
+                    # Cases like nda[nda .> 0]
+                    Tuple{BitArray},
+                    Tuple{NamedDimsArray{L,T,N,<:BitArray}} where {L,T,N},
+                    # Cases like nda[findall(f, nda)]
+                    Tuple{AbstractVector{CartesianIndex{ndims(a)}}},
+                    }
```
oxinabox: I don't like this because it isn't extensible, by other packages

mcabbott: I don't love it but... what would a better design look like?

I guess you could make `remaining_dimnames_from_indexing` return `nothing` to signal return without names?

test/wrapper_array.jl
mcabbott: These were the cases I could think of which produce a vector, not along any particular axis of the original array, and thus should drop names. Are there any others?

src/wrapper_array.jl
mcabbott: Re-done but not sure it's right yet. 

Here's another error, which could also be caught by `remaining_dimnames_from_indexing`:
```julia
ndv = NamedDimsArray(rand(1:9, 10), :x)
ndv[rand(1:10, 2,3)] # ArgumentError: A 2 dimensional array, needs a 2-tuple of dimension names. Got: (:x,)
```

src/name_core.jl
```diff
@@ -268,23 +268,40 @@ is_noninteger_type(::Any) = true
...
-@generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds)
+@generated function remaining_dimnames_from_indexing(dimnames::Tuple, inds::Tuple)
     # 0-Allocation see:
     # `@btime (()->remaining_dimnames_from_indexing((:a, :b, :c), (:,390,:)))()``
+    if length(inds.parameters) == 1 && length(dimnames.parameters) >= 2
+        type = inds.parameters[1]
+        if type <: BitArray || type <: AbstractVector{<:CartesianIndex}
+            return nothing
+        end
+    end
```
oxinabox: I think we can just make a new dispatch for this.

```
@generated function remaining_dimnames_from_indexing(
    dimnames::Tuple{<:Any, <:Any, Vararg}, inds::Tuple{T}
) where T <: Union{BitArray, AbstractVector{<:CartesianIndex}}
    return nothing
end
```


Today I learn how to use Varrg to express 2 or more
```
julia> (1,1,1) :: Tuple{Int, Int, Vararg}
(1, 1, 1)

julia> (1,1,1,1) :: Tuple{Int, Int, Vararg}
(1, 1, 1, 1)

julia> (1,1,) :: Tuple{Int, Int, Vararg}
(1, 1)

julia> (1,) :: Tuple{Int, Int, Vararg}
ERROR: TypeError: in typeassert, expected Tuple{Int64,Int64,Vararg{Any,N} where N}, got Tuple{Int64}
Stacktrace:
 [1] top-level scope at REPL[8]:1
```

src/name_core.jl
```diff
@@ -268,23 +268,40 @@ is_noninteger_type(::Any) = true
...
     keep_names = []
     dim_num = 0
     for type in inds.parameters
         if type <: Integer
```
oxinabox: Lets make a functiion that is called inside each step of this loop:
that dispatches on `type`,
and does everything? Or returns somethat that is used to do everything?

src/wrapper_array.jl
```diff
@@ -166,14 +166,21 @@ for f in (:getindex, :view, :dotview)
         end
 
         @propagate_inbounds function Base.$f(a::NamedDimsArray, inds...)
-            # Some nonscalar case, will return an array, so need to give that names.
+            # Some nonscalar case, will return an array, probably with names
             data = Base.$f(parent(a), inds...)
             L = remaining_dimnames_from_indexing(dimnames(a), inds)
-            return NamedDimsArray{L}(data)
+            if L === nothing
+                return data
```
oxinabox: Can we comment what is going on here?

src/name_core.jl
mcabbott: Hah, that's precisely what I had on the previous commit. But I guess this looked neater to me?

src/name_core.jl
mcabbott: So it returns `dim_increment, new_names` or something? I don't think you can dispatch on `ndims(type) > 1`, so there will be some duplication. Not sure.

src/wrapper_array.jl
mcabbott: Sure, will do.

src/name_core.jl
oxinabox: hmm, I guess it is something we can always change later

src/name_core.jl
oxinabox: `ndims(type)` you basically can for Tuples, and for arrays you dispatch on `AbstractVector` and then assume all others are >1.
(optionally also handle the `AbstractArray{T,0}`)

src/name_core.jl
mcabbott: OK, I pasted in your version of this, which is neater than mine was.

src/name_core.jl
mcabbott: Yea that's what I meant by duplication, it would need multiple paths to the "fall-through" case `else` here. All I've changed here is adding one more case to this list, am not so keen to re-write the rest.

src/wrapper_array.jl
```diff
@@ -166,14 +166,22 @@ for f in (:getindex, :view, :dotview)
         end
 
         @propagate_inbounds function Base.$f(a::NamedDimsArray, inds...)
-            # Some nonscalar case, will return an array, so need to give that names.
+            # Some nonscalar case, will return an array, probably with names
             data = Base.$f(parent(a), inds...)
             L = remaining_dimnames_from_indexing(dimnames(a), inds)
-            return NamedDimsArray{L}(data)
+            if L === nothing
+                # All names dropped, e.g. from mat[mat .> 0]
```
oxinabox: Why though?
I assume it is because there are no valid names for this because it has different shape to the inputs?
Comment needs to explain

src/wrapper_array.jl
```diff
@@ -166,14 +166,22 @@ for f in (:getindex, :view, :dotview)
...
+                return data
+            else
+                return NamedDimsArray{L}(data)
+            end
         end
     end
 end
 
+remaining_dimnames_from_indexing(dimnames::Tuple{Any, Any, Vararg},
+    inds::Tuple{NamedDimsArray{L,T,N,<:BitArray} where {L,T,N}}) = nothing
```
oxinabox: By BlueStyle this is not an acceptable way to do a line break
https://github.com/invenia/BlueStyle#method-definitions

test/wrapper_array.jl
```diff
@@ -72,6 +72,37 @@ end
...
+    # https://github.com/invenia/NamedDims.jl/issues/70
+    @testset "BitArray" begin
+        nda = NamedDimsArray(rand(1:9, 3,3,3), (:x, :y, :z))
+        @test dimnames(nda[rand(3) .> 0.3, 1, :]) == (:x, :z)
+
+        # flattening of ndims>1 should drop names
+        @test dimnames(nda[rand(3,3,3) .> 0.3]) == (:_,)
+        @test dimnames(nda[nda .> 3]) == (:_,)
+
+        # but vectors are unambiguous, so keep theirs
```
oxinabox: Can we break each of these, here and below into nested testsets?

```suggestion
        @testset "vectors are unambiguous, so keep their names" begin
```

src/wrapper_array.jl
mcabbott: What should it say? The output dimension isn't one of the input dimensions, as in the issue... it's not even a meaningful combination of them like `vec(x)` would be. Perhaps you have a sentence in mind?

oxinabox: I actually don't know what its doing at all, is why I want the comment

oxinabox: Oh after thinking a bit i see there is no other way for it to be.
Maybe we can say something like: 

> `mat[mat .> 0.5] is the same as `vec(mat)[vec(mat .> 0.5)]` and so ravels the n dimension down to one new combined one.

test/wrapper_array.jl
```diff
@@ -72,6 +72,37 @@ end
         @test size(nda[CartesianIndex(1, 1), newaxis]) == (1, )
         @test nda[CartesianIndex(1, 1), newaxis][1] == nda[1, 1]
     end
+
+    # https://github.com/invenia/NamedDims.jl/issues/70
+    @testset "BitArray" begin
```
oxinabox: should we replicate this test set, for `Array{Bool}` which should act the same?
```
julia> x = rand(3,2)
3×2 Array{Float64,2}:
 0.241709  0.691836
 0.671663  0.258613
 0.705006  0.366028

julia> x[x .> 0.5]
3-element Array{Float64,1}:
 0.6716626129087626
 0.7050058537208825
 0.6918362123847104

julia> collect(x .> 0.5)
3×2 Array{Bool,2}:
 0  1
 1  0
 1  0


julia> x[collect(x .> 0.5)]
3-element Array{Float64,1}:
 0.6716626129087626
 0.7050058537208825
 0.6918362123847104

```

src/wrapper_array.jl
mcabbott: ok, added that a few lines down

test/wrapper_array.jl
mcabbott: Good idea, it didn't work before, but now does & is tested.

mcabbott:
I wonder whether `nothing` is the right flag here. Notice this:
```
NamedDims.remaining_dimnames_from_indexing((:a, :b), (1,2)) == ()
NamedDims.remaining_dimnames_from_indexing((:a, :b), (CartesianIndex(1,2),)) == ()
```
Of course these cases are caught by special methods right now. But perhaps they should be united with the `BitArray` etc. cases, with `()` as a flag?

Also, aside, the test `@test 0 == @allocated NamedDimsArray(cnda, (:x, :_))` fails on 1.4, I don't know why. Travis is only run on versions up to 1.3 right now though.

mcabbott:
See what you think of this last commit. It's not the minimal change. But perhaps something like this is cleaner than bolting on exceptions.

src/wrapper_array.jl
```diff
@@ -155,24 +155,12 @@ for f in (:getindex, :view, :dotview)
...
-
-        @propagate_inbounds function Base.$f(a::NamedDimsArray, ci::CartesianIndex)
-            # Easy scalar case, will just return the element
-            return Base.$f(parent(a), ci)
-        end
-
-        @propagate_inbounds function Base.$f(a::NamedDimsArray, inds...)
-            # Some nonscalar case, will return an array, probably with names
+        @propagate_inbounds function Base.$f(a::NamedDimsArray, raw_inds...)
+            inds = Base.to_indices(a, raw_inds)
```
oxinabox: can I summon @iamed2  for review here.
They understand `to_indices` better than I do, infact when I wrote this code I didn't know it existed.

 think this is a good idea and we should ghave been doing it this way all along, but not 100% sure

src/wrapper_array.jl
```diff
@@ -155,24 +155,12 @@ for f in (:getindex, :view, :dotview)
...
             data = Base.$f(parent(a), inds...)
             L = remaining_dimnames_from_indexing(dimnames(a), inds)
             if L === nothing
-                # This is a signal to drop all names, e.g.
-                # mat[mat .> 0.5] returns a vector, the same as vec(mat)[vec(mat .> 0.5)],
-                # whose dimension isn't any of the original dimensions, hence has no names.
+                # Case of scalar output, plus some weird cases like mat[mat .> 0]
```
oxinabox:
```suggestion
                # Case of scalar output, or cases that merge dimensions down to vector like `mat[mat .> 0]`
```

src/wrapper_array.jl
```diff
@@ -155,21 +155,16 @@ for f in (:getindex, :view, :dotview)
...
-        @propagate_inbounds function Base.$f(a::NamedDimsArray, ci::CartesianIndex)
-            # Easy scalar case, will just return the element
-            return Base.$f(parent(a), ci)
-        end
-
-        @propagate_inbounds function Base.$f(a::NamedDimsArray, inds...)
-            # Some nonscalar case, will return an array, so need to give that names.
+        @propagate_inbounds function Base.$f(a::NamedDimsArray, raw_inds...)
+            inds = Base.to_indices(a, raw_inds)
             data = Base.$f(parent(a), inds...)
```
mcabbott: It is tempting here to insert
```julia
data isa AbstractArray || return data
```
and thus only run `remaining_dimnames_from_indexing` at all when the answer is not a scalar. Although it should be compiled away, I think, and this goes back to having two paths which return `data`. 

BTW, the speed of indexing like `@btime $nda[1,1]` is unaffected by this PR.

src/name_core.jl
mcabbott: If we were to re-write this function, I suspect it should be done by recursively peeling elements off these two tuples. It probably need not be generated. But I have not done this.

src/wrapper_array.jl
mcabbott: Not en expert, but I start to think this is the correct official thing to do here. For instance it will mean that https://github.com/mbauman/InvertedIndices.jl works by design, rather than by good luck.

iamed2: I believe that if you're passing through to `getindex` or `view` (`dotview` passes through to either `getindex` or `view` so I'm not sure why you're defining `dotview` here), you can expect `to_indices` to be called for you. I would not call `to_indices` here.

iamed2: `dotview` issue spun off here: https://github.com/invenia/NamedDims.jl/pull/109

src/wrapper_array.jl
```diff
@@ -181,6 +176,7 @@ end
     return setindex!(a, value, inds...)
 end
 
-@propagate_inbounds function Base.setindex!(a::NamedDimsArray, value, inds...)
+@propagate_inbounds function Base.setindex!(a::NamedDimsArray, value, raw_inds...)
+    inds = Base.to_indices(a, raw_inds)
```
iamed2: I don't believe this is necessary. Can you provide a case where this succeeds and the previous version fails?

mcabbott: I agree, I think I unthinkingly made them the same. Will remove.

src/wrapper_array.jl
mcabbott: The point of doing this for `getindex` is that weird index types this package doesn't know about are then normalised, and can be safely handled by `remaining_dimnames_from_indexing`.

(No opinion on `dotview`, it was there a long time.)

mcabbott: Here's an example of what I mean about normalisation:
```julia
using NamedDims, InvertedIndices
nda = NamedDimsArray([10 20; 30 40], (:x, :y));
dimnames(nda[Not(CartesianIndex(1,1))]) # (:x,) on tagged version, accidentally

typeof(Not(CartesianIndex(1,1))) # InvertedIndex{CartesianIndex{2}}
supertype(ans) # Any

inds = Base.to_indices(nda, (Not(CartesianIndex(1,1)),))
supertype(typeof(inds[1])) # AbstractArray{CartesianIndex{2},1}
```

iamed2: You then lose the behaviour of any `getindex` that _doesn't_ just depend on `to_indices`. I was worried about that but then I decided those array types should just fix themselves.

So this is probably fine then. Normalizing for `remaining_dimnames_from_indexing` makes sense.

mcabbott: Great, then perhaps this is done.

oxinabox:
Is this ready to merge?

mcabbott:
Yes I think so.

oxinabox:
thanks!

mcabbott:
This one really took on a life of its own, but I learned some things!

PR #105, opened 2020-03-17
Simpler handling of transpose(transpose()) etc.

mcabbott:
This implements the "maximal fix" suggested in https://github.com/invenia/NamedDims.jl/pull/77, which reads the dimension actually created by `transpose(data)` rather than predicting this from the type `Transpose{T,<:Vector}` etc. 

The reason to do this is that, if another wrapper behaves like `NamedDimsArray` does, and wraps such a `Transpose`, then it can't at present live inside this one. In particular, I want this for commutative wrappers here:

https://github.com/mcabbott/AxisKeys.jl/blob/3a3907afd475098e499188715809f5f2a2c28b8c/test/_functions.jl#L35

### DIFF
--- a/src/functions_dims.jl
+++ b/src/functions_dims.jl
@@ -24,27 +24,18 @@ for f in (
     :(Base.permutedims),
     :(LinearAlgebra.pinv)
 )
-    # Vector
-    @eval function $f(nda::NamedDimsArray{L,T,1}) where {L,T}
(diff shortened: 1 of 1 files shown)


oxinabox:
I think this makes sense.

mcabbott:
I bumped the version in #106, perhaps they can share.

PR #114, opened 2020-07-17
Update .travis.yml

mzgubic:
@nickrobinson251 is this the change you meant? Sorry if it's wrong, first time doing something to .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,8 @@ os:
   - linux
   - osx
 julia:
-  - 1.0
-  - 1.1
-  - 1.2
-  - 1.3
-  - 1.4
-  - 1.5
+  - 1.0 # long-term support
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,6 +8,8 @@ julia:
   - 1.1
   - 1.2
   - 1.3
+  - 1.4
+  - 1.5
```
nickrobinson251, line 12: Thanks! These are exactly the right lines to change, but I don't think we need to test every version, and to save us having to add new versions in the future we can just use `1` which will get us the latest stable  `1.x` version (so 1.4 right now, but 1.5 soon)
```
julia:
  - 1.0   # long-term support version
  - 1     # latest stable version
  - nightly
```

Issue #60, opened 2019-07-13
Method Ambiguous Error *(::Adjoint{Float64,Array{Float64,2}}, ::NamedDimsArray


AlexRobson:
MFE:
```
d = NamedDimsArray{(:features, :observations)}(rand(4, 10))
β = rand(4, 3)
β' * d[:, 1]
MethodError: *(::Adjoint{Float64,Array{Float64,2}}, ::NamedDimsArray{(:features,),Float64,1,Array{Float64,1}}) is ambiguous. Candidates:
  *(a::AbstractArray{T,2} where T, b::NamedDimsArray{#s19,#s20,1,A} where A<:AbstractArray{#s20,1} where #s20 where #s19) in NamedDims at /Users/alexr/.julia/packages/NamedDims/MjTg0/src/functions_math.jl:74
  *(adjA::Adjoint{#s623,#s622} where #s622<:AbstractArray{T,2} where #s623, x::AbstractArray{S,1}) where {T, S} in LinearAlgebra at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/LinearAlgebra/src/matmul.jl:103
Possible fix, define
  *(::Adjoint{#s623,#s622} where #s622<:AbstractArray{T,2} where #s623, ::NamedDimsArray{#s19,S,1,A} where A<:AbstractArray{S,1} where #s19)
```

Where as this is fine:
```
d = NamedDimsArray{(:features, :observations)}(rand(4, 10))
β = rand(4, 3)
collect(β') * d[:, 1]
```

oxinabox:
Something is  missing somewhere in here
https://github.com/invenia/NamedDims.jl/blob/master/src/functions_math.jl

mzgubic:
closed by #118

PR #118, opened 2020-07-20
Resolve method ambiguity on Adjoint and Diagonal

### LINKED ISSUE #112: Ambiguity on 1.5: Adjoint{...,Vector} * NamedDimsArray{...,Vector}
Julia 1.5 seems to cause extra method ambiguities, for instance this:
```julia
julia> using NamedDims

julia> rand(3)' * NamedDimsArray(rand(3), :a)
ERROR: MethodError: *(::LinearAlgebra.Adjoint{Float64,Array{Float64,1}}, ::NamedDimsArray{(:a,),Float64,1,Array{Float64,1}}) is ambiguous. Candidates:
  *(u::LinearAlgebra.Adjoint{T,var"#s814"} where var"#s814"<:(AbstractArray{T,1} where T), v::AbstractArray{T,1}) where T<:Number in LinearAlgebra at /Applications/Julia-1.5.app/Contents/Resources/julia/share/julia/stdlib/v1.5/LinearAlgebra/src/adjtrans.jl:267
  *(a::LinearAlgebra.Adjoint{var"#s16",var"#s15"} where var"#s15"<:(AbstractArray{T,1} where T) where var"#s16", b::NamedDimsArray{L,T,1,var"#s17"} where var"#s17"<:AbstractArray{T,1}) where {L, T} in NamedDims at /Users/me/.julia/packages/NamedDims/p7e4G/src/functions_math.jl:48
Possible fix, define
  *(::LinearAlgebra.Adjoint{T,var"#s15"} where var"#s15"<:(AbstractArray{T,1} where T), ::NamedDimsArray{L,T,1,var"#s17"} where var"#s17"<:AbstractArray{T,1}) where {L, T<:Number}
Stacktrace:
 [1] top-level scope at REPL[2]:1

julia> [ones(2,2), ones(2,2)]' * NamedDimsArray([ones(2,2), ones(2,2)], :a)
2×2 Array{Float64,2}:
 4.0  4.0
 4.0  4.0
```

xref https://github.com/mcabbott/AxisKeys.jl/issues/17 

Edit: also here: https://github.com/JuliaCI/NanosoldierReports/blob/master/pkgeval/by_date/2020-05/09/logs/NamedDims/1.6.0-DEV-b828bbc426.log#L129

mzgubic:
Adds tests and the two methods which resolve the ambiguities. Closes #112 and #60.

### DIFF
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -83,7 +83,7 @@ end
 # but that overwrites existing *(1D NDA, Vector) methods
 # should improve the macro above to deal with this case
 function Base.:*(a::Diagonal, b::NamedDimsArray{<:Any, <:Any, 1})
-    return *(NamedDims.NamedDimsArray{dimnames(a)}(a), b)
+    return *(NamedDimsArray{dimnames(a)}(a), b)
 end
 
 function Base.:*(a::NamedDimsArray{<:Any, <:Any, 1}, b::Diagonal)
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -78,9 +78,20 @@ macro declare_matmul(MatrixT, VectorT=nothing)
     return esc(Expr(:block, codes...))
 end
 
+# The following two methods can be defined by using
+# @declare_matmul(Diagonal, AbstractVector)
+# but that overwrites existing *(1D NDA, Vector) methods
+# should improve the macro above to deal with this case
+function Base.:*(a::Diagonal, b::NamedDimsArray{<:Any, <:Any, 1})
+    return *(NamedDims.NamedDimsArray{dimnames(a)}(a), b)
+end
+
+function Base.:*(a::NamedDimsArray{<:Any, <:Any, 1}, b::Diagonal)
+    return *(a, NamedDims.NamedDimsArray{dimnames(b)}(b))
+end
+
 @declare_matmul(AbstractMatrix, AbstractVector)
 @declare_matmul(Adjoint{<:Any, <:AbstractMatrix{T1}} where T1, Adjoint{<:Any, <:AbstractVector})
-@declare_matmul(Diagonal, AbstractVector)
 @declare_matmul(Diagonal,)
 
 function Base.inv(nda::NamedDimsArray{L, T, 2}) where {L,T}
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -43,8 +43,12 @@ function Base.:*(a::NamedDimsArray{L,T,2,<:CoVector}, b::AbstractVector) where {
     return *(parent(a), b)
 end
 
-# Using `CovVector` results in Method ambiguities; have to define more specific methods.
-for A in (Adjoint{<:Any, <:AbstractVector}, Transpose{<:Real, <:AbstractVector{<:Real}})
+# Using `CoVector` results in Method ambiguities; have to define more specific methods.
+specifics = (Adjoint{<:Any, <:AbstractVector},
+             Adjoint{<:Any, <:AbstractMatrix{T1}} where T1,
+             Transpose{<:Real, <:AbstractVector{<:Real}},
+             Diagonal{<:Any})
+for A in specifics
     @eval function Base.:*(a::$A, b::NamedDimsArray{L,T,1,<:AbstractVector{T}}) where {L,T}
         return *(a, parent(b))
     end
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -43,12 +43,8 @@ function Base.:*(a::NamedDimsArray{L,T,2,<:CoVector}, b::AbstractVector) where {
     return *(parent(a), b)
 end
 
-# Using `CoVector` results in Method ambiguities; have to define more specific methods.
-specifics = (Adjoint{<:Any, <:AbstractVector},
-             Adjoint{<:Any, <:AbstractMatrix{T1}} where T1,
-             Transpose{<:Real, <:AbstractVector{<:Real}},
-             Diagonal{<:Any})
-for A in specifics
+# Using `CovVector` results in Method ambiguities; have to define more specific methods.
+for A in (Adjoint{<:Any, <:AbstractVector}, Transpose{<:Real, <:AbstractVector{<:Real}})
     @eval function Base.:*(a::$A, b::NamedDimsArray{L,T,1,<:AbstractVector{T}}) where {L,T}
         return *(a, parent(b))
     end
--- a/src/functions_math.jl
+++ b/src/functions_math.jl
@@ -83,6 +83,8 @@ macro declare_matmul(MatrixT, VectorT=nothing)
 end
 
 @declare_matmul(AbstractMatrix, AbstractVector)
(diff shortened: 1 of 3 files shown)


src/functions_math.jl
```diff
@@ -43,8 +43,12 @@ function Base.:*(a::NamedDimsArray{L,T,2,<:CoVector}, b::AbstractVector) where {
...
 end
 
-# Using `CovVector` results in Method ambiguities; have to define more specific methods.
-for A in (Adjoint{<:Any, <:AbstractVector}, Transpose{<:Real, <:AbstractVector{<:Real}})
+# Using `CoVector` results in Method ambiguities; have to define more specific methods.
+specifics = (Adjoint{<:Any, <:AbstractVector},
+             Adjoint{<:Any, <:AbstractMatrix{T1}} where T1,
+             Transpose{<:Real, <:AbstractVector{<:Real}},
+             Diagonal{<:Any})
+for A in specifics
```
oxinabox, line 51: Two things are unclear to me:

### Why is this the right way to handle `Diagonal` rather than `@declare_matmul` below ?
In particular, because the others in this loop perform noperation that returns a scalar, and thus doesn't need to worry about preserving names, where as for `Diagonal`  that is not the case.

```julia
julia> [10, 20, 30]' * [1,2,3]
140
```
vs
```julia
julia> Diagonal([1,2]) * [10, 20]
2-element Array{Int64,1}:
 10
 40
```


### why is this the right way to handle the product of the `Adjoint` of a Matrix with a Vector

Which is again is an operation that does does not return scalar and thus needs to preserve names.
```julia
julia> [10 20; 31 32]' * [1,2]
2-element Array{Int64,1}:
 72
 84
```

mzgubic: They were indeed not the right way, thanks for pointing it out. I've added tests which ensure that NamedDimsArray is returned from these operations, rather than an array which was the case in the first implementation.

src/functions_math.jl
```diff
@@ -83,6 +83,8 @@ macro declare_matmul(MatrixT, VectorT=nothing)
 end
 
 @declare_matmul(AbstractMatrix, AbstractVector)
+@declare_matmul(Adjoint{<:Any, <:AbstractMatrix{T1}} where T1, Adjoint{<:Any, <:AbstractVector})
+@declare_matmul(Diagonal, AbstractVector)
```
mzgubic, line 83: @oxinabox ideas here?

Adding `@declare_matmul(Diagonal, AbstractVector)` is kind of awkward, because AbstractVector isn't a 1D version of Diagonal. What happens is that certain methods get overwritten, for example
`*(Vector, 2D NDA)`

However it is needed in order to cover `*(Diagonal, 1D NDA)` case.

oxinabox: maybe `@declare_matmul` needs to be modified so that
`@declare_matmul(Diagonal)` does the right thing.

mzgubic: Hmm, ok, that's probably beyond my current knowledge of Julia. I could hack something but it wouldn't be pretty.
Maybe we should just remove L87 and open a new issue?

oxinabox: Can you just write out the methods you want to create directly and explictly?

That will solve it, and maybe later we can refactor it to be done in the macro.

mzgubic: I've added the two methods manually. Once this PR is merged I will open an issue highlighting this.

src/functions_math.jl
```diff
@@ -78,7 +78,20 @@ macro declare_matmul(MatrixT, VectorT=nothing)
     return esc(Expr(:block, codes...))
 end
 
+# The following two methods can be defined by using
+# @declare_matmul(Diagonal, AbstractVector)
+# but that overwrites existing *(1D NDA, Vector) methods
+# should improve the macro above to deal with this case
+function Base.:*(a::Diagonal, b::NamedDimsArray{<:Any, <:Any, 1})
+    return *(NamedDims.NamedDimsArray{dimnames(a)}(a), b)
```
nickrobinson251, line 86:
```suggestion
    return *(NamedDimsArray{dimnames(a)}(a), b)
```

src/functions_math.jl
```diff
@@ -78,7 +78,20 @@ macro declare_matmul(MatrixT, VectorT=nothing)
...
+end
+
+function Base.:*(a::NamedDimsArray{<:Any, <:Any, 1}, b::Diagonal)
+    return *(a, NamedDims.NamedDimsArray{dimnames(b)}(b))
```
nickrobinson251, line 90:
```suggestion
    return *(a, NamedDimsArray{dimnames(b)}(b))
```

src/functions_math.jl
```diff
@@ -78,7 +78,20 @@ macro declare_matmul(MatrixT, VectorT=nothing)
     return esc(Expr(:block, codes...))
 end
 
+# The following two methods can be defined by using
+# @declare_matmul(Diagonal, AbstractVector)
+# but that overwrites existing *(1D NDA, Vector) methods
+# should improve the macro above to deal with this case
+function Base.:*(a::Diagonal, b::NamedDimsArray{<:Any, <:Any, 1})
+    return *(NamedDims.NamedDimsArray{dimnames(a)}(a), b)
+end
```
nickrobinson251, line 87: wrapping a Diagonal in a NDA does seem a roundabout way to do this... i don't have a better proposal, but it makes me suspect something is still not quite right somewhere

src/functions_math.jl
```diff
@@ -78,7 +78,20 @@ macro declare_matmul(MatrixT, VectorT=nothing)
...
+
+function Base.:*(a::NamedDimsArray{<:Any, <:Any, 1}, b::Diagonal)
+    return *(a, NamedDims.NamedDimsArray{dimnames(b)}(b))
+end
+
 @declare_matmul(AbstractMatrix, AbstractVector)
+@declare_matmul(Adjoint{<:Any, <:AbstractMatrix{T1}} where T1, Adjoint{<:Any, <:AbstractVector})
```
nickrobinson251, line 94: is the `where T1` needed to break the ambiguity?
```suggestion
@declare_matmul(Adjoint{<:Any, <:AbstractMatrix}, Adjoint{<:Any, <:AbstractVector})
```

src/functions_math.jl
mzgubic: This is what the `declare_matmul` macro does as well. I think the idea is to reuse the NDA * NDA method, which deals with the naming of output dimensions. I guess you are worried about the conversion overhead?

nickrobinson251: No, there shouldn't be an overhead - I was his surprised by the approach

src/functions_math.jl
mzgubic: Yes, I believe it is needed.

`AbstractMatrix` is an alias for 'AbstractArray{T1, 2}, where T1', which puts the where T1 inside the outer curly brace.

I tried really understanding why this is needed (https://stackoverflow.com/questions/62995469/difference-between-where-keyword-inside-or-outside-the-curly-braces), but to be honest I still don't quite get why the particular form is needed here..

nickrobinson251: That was the we have to be so specific is such a pain (again, makes me worry we'll be playing method-ambiguity-whack-a-mole forever)

Anyway, thanks for solving this particularly issue - a good one to solve

PR #119, opened 2020-07-21
Fix the getindex/setindex bug referenced in #110

### LINKED ISSUE #110: getindex with no arguments returns the whole array
I'm on latest release (0.2.21) and have noticed the following inconsistency between named arrays and normal arrays:

```julia
julia> using NamedDims

julia> s = NamedDimsArray{(:a,)}([1])
1-element NamedDimsArray{(:a,),Int64,1,Array{Int64,1}}:
 1

julia> s[]
1-element NamedDimsArray{(:a,),Int64,1,Array{Int64,1}}:
 1

julia> t = [1]
1-element Array{Int64,1}:
 1

julia> t[]
1
```

That is to say, `getindex(v::Vector)` returns the unique element if the vector has length one, and errors otherwise. OTOH, `getindex(x::NamedDimsArray) == x`. For 0-dimensional arrays, where `getindex(v)` should return the unique entry, one gets a stack overflow error with `NamedDimsArray{T, 0}`:

```julia
julia> s = NamedDimsArray{()}(fill(1));

julia> s[]
ERROR: StackOverflowError:
Stacktrace:
 [1] getindex(::NamedDimsArray{(),Int64,0,Array{Int64,0}}; named_inds::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/pietro/.julia/packages/NamedDims/p7e4G/src/wrapper_array.jl:155
 [2] getindex(::NamedDimsArray{(),Int64,0,Array{Int64,0}}) at /home/pietro/.julia/packages/NamedDims/p7e4G/src/wrapper_array.jl:154
 ... (the last 2 lines are repeated 39990 more times)
 [79983] getindex(::NamedDimsArray{(),Int64,0,Array{Int64,0}}; named_inds::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/pietro/.julia/packages/NamedDims/p7e4G/src/wrapper_array.jl:155
```

mzgubic:
Closes #110.

### DIFF
--- a/test/wrapper_array.jl
+++ b/test/wrapper_array.jl
@@ -54,13 +54,15 @@ end
 
     @testset "No arguments" begin
         ndm = NamedDimsArray([10 20; 30 40], (:x, :y))
-        ndv = NamedDimsArray([1, 2, 3], (:x,))
         @test_throws BoundsError ndm[]
+
+        ndv = NamedDimsArray([1, 2, 3], (:x,))
         @test_throws BoundsError ndv[]
 
         nds = NamedDimsArray([4], (:x,))  # 1d 1el array (vector with one element)
-        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (scalar)
         @test nds[] == 4
+
+        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (kind of like a scalar)
         @test nds2[] == 4
     end
 
@@ -157,9 +159,12 @@ end
 
     @testset "no arguments" begin
         nds = NamedDimsArray([4], (:x,))  # 1d 1el array (vector with one element)
-        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (kind of like a scalar)
         @test (nds[] = 2) == 2
+        @test nds[] == 2
+
+        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (kind of like a scalar)
         @test (nds2[] = 2) == 2
+        @test nds2[] == 2
     end
 
     @testset "by position" begin
--- a/test/wrapper_array.jl
+++ b/test/wrapper_array.jl
@@ -52,6 +52,18 @@ end
         @test view(nda, 1,1) == view(nda.data, 1,1)
     end
 
+    @testset "No arguments" begin
+        ndm = NamedDimsArray([10 20; 30 40], (:x, :y))
+        ndv = NamedDimsArray([1, 2, 3], (:x,))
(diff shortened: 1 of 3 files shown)


test/wrapper_array.jl
```diff
@@ -52,6 +52,18 @@ end
...
+    @testset "No arguments" begin
+        ndm = NamedDimsArray([10 20; 30 40], (:x, :y))
+        ndv = NamedDimsArray([1, 2, 3], (:x,))
+        @test_throws BoundsError ndm[]
+        @test_throws BoundsError ndv[]
+
+        nds = NamedDimsArray([4], (:x,))  # 1d 1el array (vector with one element)
+        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (scalar)
+        @test nds[] == 4
+        @test nds2[] == 4
```
oxinabox, line 64: The general pattern for tests is:
*arrange-act-assert*, **arrange-act-assert*, *arrange-act-assert*.
Rather than:
**arrange-arrange-arrange, act-act-act, assert-assert-assert**
so as to minimize the distances between the setup and the assertion.

some times we violate that a bit by declaring a dome things at the start and reusing them for many assertions.
but here we use each dclared variable once.
So we can do declare-test declare-test
and minimise distance we need to do that context over

test/wrapper_array.jl
```diff
@@ -143,6 +155,13 @@ end
         @test nda == [1000 1000; 30 40]
     end
 
+    @testset "no arguments" begin
+        nds = NamedDimsArray([4], (:x,))  # 1d 1el array (vector with one element)
+        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (scalar)
```
oxinabox, line 160:
```suggestion
        nds2 = NamedDimsArray{()}(fill(4)); # 0d 1el array (kind of like a scalar)
```

test/wrapper_array.jl
```diff
@@ -143,6 +155,13 @@ end
...
+        @test (nds[] = 2) == 2
+        @test (nds2[] = 2) == 2
```
oxinabox, line 162: This seems like it is only testing the return value that is returned by `setindex`
we should also be testing what the value of `nds` and `nds2` are after this operation.

test/wrapper_array.jl
mzgubic: Thanks for the added explanation, these are really helpful!

mzgubic:
good to go?

PR #122, opened 2020-07-28
Extend CovarianceEstimation.cov

### LINKED ISSUE #120: Extend CovarianceEstimation.cov
The current covariance (and correlation) methods overload `Statistics.cov` (`Statistics.cor`).

When using [CovarianceEstimation.jl](https://github.com/mateuszbaran/CovarianceEstimation.jl) another `cov` method is exported that takes a `CovarianceEstimator` type.

The issue is that as of 0.2.4, `cov` returns a `Symmetric` matrix which incorrectly wraps the `NamedDimsArray` if it gets given one.

```julia
julia> using CovarianceEstimation, LinearAlgebra, NamedDims

julia> x = NamedDimsArray{(:a, :b)}(rand(5, 5));

julia> estimator = LinearShrinkage(DiagonalCommonVariance(); corrected=true)

julia> cov(estimator, x)
5×5 Symmetric{Float64,NamedDimsArray{(:b, :b),Float64,2,Array{Float64,2}}}:
  0.0668947  -0.0         0.0         0.0        -0.0
 -0.0         0.0668947  -0.0         0.0         0.0
  0.0        -0.0         0.0668947  -0.0         0.0
  0.0         0.0        -0.0         0.0668947  -0.0
 -0.0         0.0         0.0        -0.0         0.0668947
```

We need to extend the function properly with:
```julia
function CovarianceEstimation.cov(
    estimator::LinearShrinkage, a::NamedDimsArray{L, T, 2};
    dims=1,
) where {L, T}
    numerical_dims = dim(a, dims)
    data = cov(estimator, parent(a); dims=numerical_dims)
    names = symmetric_names(L, numerical_dims)
    return NamedDimsArray{names}(data)
end
```

then we get
```julia
julia> cov(LinearShrinkage(DiagonalCommonVariance(); corrected=true), x)
5×5 NamedDimsArray{(:b, :b),Float64,2,Symmetric{Float64,Array{Float64,2}}}:
  0.0668947  -0.0         0.0         0.0        -0.0
 -0.0         0.0668947  -0.0         0.0         0.0
  0.0        -0.0         0.0668947  -0.0         0.0
  0.0         0.0        -0.0         0.0668947  -0.0
 -0.0         0.0         0.0        -0.0         0.0668947
```

glennmoy:
Closes #120

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["Invenia Technical Computing Corporation"]
 version = "0.2.24"
 
 [deps]
-CovarianceEstimation = "587fd27a-f159-11e8-2dae-1979310e6154"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
@@ -19,9 +18,10 @@ julia = "1"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+CovarianceEstimation = "587fd27a-f159-11e8-2dae-1979310e6154"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 
 [targets]
-test = ["BenchmarkTools", "SparseArrays", "Test", "Tracker"]
+test = ["BenchmarkTools", "CovarianceEstimation", "SparseArrays", "Test", "Tracker"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,9 +1,10 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.23"
+version = "0.2.24"
 
 [deps]
+CovarianceEstimation = "587fd27a-f159-11e8-2dae-1979310e6154"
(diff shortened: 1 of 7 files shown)


oxinabox:
This should be done via REQUIRES.jl, same as we do Tracker.

glennmoy:
> This should be done via REQUIRES.jl, same as we do Tracker.

ah, interesting - thanks!

glennmoy:
I've been getting the following warning but can't seem to figure out why. I'm `using CovarianceEstimation` the same as `Tracker`, which doesn't throw a warning.

```julia
┌ Warning: Package NamedDims does not have CovarianceEstimation in its dependencies:
│ - If you have NamedDims checked out for development and have
│   added CovarianceEstimation as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with NamedDims
└ Loading CovarianceEstimation into NamedDims from project dependency, future warnings for NamedDims are suppressed.
Test Summary: | Pass  Broken  Total
NamedDims.jl  |  896       3    899
    Testing NamedDims tests passed 
```

mzgubic:
I found this:
https://github.com/JuliaPackaging/Requires.jl/issues/52

So you can either add a `.` to `using CovarianceEstimation` to make it `using .CovarianceEstimation`

or delete the line and do
```
for E in (:(CovarianceEstimation.LinearShrinkage),
          :(CovarianceEstimation.SimpleCovariance),
          :(CovarianceEstimation.AnalyticalNonlinearShrinkage))
```

rofinn:
FWIW, I'm not sure I like using Requires.jl for this. If the `cov` api changes at any point then we can't handle that disconnect. That seems particularly likely given that the package is only on 0.2.5.

oxinabox:
OK, I am sold, lets extend it directly for now

oxinabox:
For now let's just add a direct dependency, and change it later.
This is blocking us getting performance enhancements of more recent CovarianceEstimstion releases

mzgubic:
rebased in https://github.com/invenia/NamedDims.jl/pull/174

PR #124, opened 2020-08-06
Implement cat, vcat, hcat

### LINKED ISSUE #123: cat, hcat, vcat only work with numbered dimensions
I guess this could be added relatively easily? Would appreciate guidance but happy to give it a go.

```
julia> nda = NamedDimsArray(zeros(2, 5), (:x, :y))
2×5 NamedDimsArray{(:x, :y),Float64,2,Array{Float64,2}}:
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0

julia> cat(nda, ones(1, 5), dims=1)
3×5 NamedDimsArray{(:x, :y),Float64,2,Array{Float64,2}}:
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 1.0  1.0  1.0  1.0  1.0

julia> cat(nda, ones(1, 5), dims=x)
ERROR: UndefVarError: x not defined
...

julia> cat(nda, ones(1, 5), dims=:x)
ERROR: MethodError: no method matching isless(::Symbol, ::Int64)
...
```

mzgubic:
Once completed, closes #123 

Here's my first attempt at the problem. The current state has a few improvements, namely:
1) Returns NamedDimsArray with correct dimensions
2) Errors when dimensions don't match
3) Can call cat with named dimensions

However, there are a few things that should be polished:
A) ~~Calling cat with `dims=:new` where `:new` is a new dimension errors.~~
B) ~~`cat` method is only defined for cases where there are two arrays (at least one of which is an NDA)~~
C) ~~Current naive hcat and vcat definitions create ambiguities, I think also because the Base methods accept arbitrary number of arguments.~~

Details on 1:
```
julia> a = [10 20; 31 40];

julia> nda = NamedDimsArray(a, (:x, :y));

julia> cat(nda, nda, dims=3)
2×2×2 NamedDimsArray{(:x, :y, :_),Int64,3,Array{Int64,3}}:
[:, :, 1] =
 10  20
 31  40

[:, :, 2] =
 10  20
 31  40

julia> cat(nda, nda, dims=:z)
ERROR: ArgumentError: Specified name (:z) does not match any dimension name ((:x, :y))
```

Details on 2:
```
julia> cat(nda, nda, nda', dims=2) # should error
2×6 NamedDimsArray{(:x, :y),Int64,2,Array{Int64,2}}:
 10  20  10  20  10  31
 31  40  31  40  20  40
```

Details on 3:
```
julia> hcat(nda, a)
ERROR: MethodError: hcat(::NamedDimsArray{(:x, :y),Int64,2,Array{Int64,2}}, ::Array{Int64,2}) is ambiguous. Candidates:
  hcat(A::Union{AbstractArray{T,1}, AbstractArray{T,2}}...) where T in Base at abstractarray.jl:1313
  hcat(a::NamedDimsArray, b::AbstractArray) in NamedDims at /Users/mzgubic/.julia/dev/NamedDims.jl/src/functions.jl:179
  hcat(A::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T...) in Base at abstractarray.jl:1312
Possible fix, define
  hcat(::Union{NamedDimsArray{L,T,1,A} where A<:AbstractArray{T,1} where L, NamedDimsArray{L,T,1,A} where A<:AbstractArray{T,1} where L, NamedDimsArray{L,T,2,A} where A<:AbstractArray{T,2} where L, NamedDimsArray{L,T,2,A} where A<:AbstractArray{T,2} where L}, ::Union{AbstractArray{T,1}, AbstractArray{T,2}, AbstractArray{T,1}, AbstractArray{T,2}}) where T
Stacktrace:
 [1] top-level scope at REPL[186]:1
 [2] eval(::Module, ::Any) at ./boot.jl:331
 [3] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [4] run_backend(::REPL.REPLBackend) at /Users/mzgubic/.julia/packages/Revise/BqeJF/src/Revise.jl:1184
 [5] top-level scope at none:0
```


src/functions.jl
```diff
@@ -146,6 +146,54 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+function Base.cat(a::NamedDimsArray{La}, b::NamedDimsArray{Lb}; dims) where {La, Lb}
+    newL = unify_names_shortest(La, Lb)
+    numerical_dims = dim(newL, dims)
+    data = Base.cat(parent(a), parent(b); dims=numerical_dims)
+    newL = unify_names_longest(newL, dimnames(data)) # when dims=3 for two 2d arrays
+    return NamedDimsArray{newL}(data)
+end
+
+function Base.cat(a::NamedDimsArray, b::AbstractArray...; dims)
+    return Base.cat(a, Base.cat(b...; dims=dims); dims=dims)
```
mzgubic, line 175: I really dislike this because it only works if the first argument is NDA. If the first argument is a regular array and any of the other ones is NDA it doesn't work. Any suggestions on a pattern that solves this?

I've asked here: https://stackoverflow.com/questions/63285604/how-to-dispatch-based-on-the-type-of-any-of-the-splattered-args

mcabbott: One way `@eval` a fairly long list of possibilities... I guess just `nda, abstractarray`, `a, nda`, `nda,nda` will allow it to be triggered if any of the first two are named. (If you want any-of-first-three, then it's a longer list.) 

Perhaps useful:

https://github.com/mcabbott/AxisKeys.jl/blob/master/src/functions.jl#L93

mzgubic: Thanks! It now works if any of the first two args is an NDA.

mzgubic:
Does defining `expand_dimnames` look like a reasonable solution to solving A)? Should be self contained in this commit: https://github.com/invenia/NamedDims.jl/pull/124/commits/2aef8e539858ca616d2da41b334163af6f427314

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
 
+################################################
+# cat, vcat, hcat
+
+function Base.cat(a::NamedDimsArray{L}; dims) where L
+    newL = expand_dimnames(L, dims)
+    numerical_dims = dim(newL, dims)
+    data = Base.cat(parent(a); dims=numerical_dims)
+    newL = unify_names_longest(newL, dimnames(data))
+    return NamedDimsArray{newL}(data)
```
oxinabox, line 157: reusing the name `newL` for variables of different types (in this case tuples of different lengths),
leads to type-instability.

I am not sure if there is an easier way to see this than using Cthulu.jl,
and viewing with `[w]arn` since it is a kwarg and so `@code_warntype` will only show the keyword method being called.
I guess you could rewrite it not to use a kwarg just for testing this out.

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+for (T, S) in [
+    (:NamedDimsArray, :AbsVecOrMat),
+    (:NamedDimsArray, :AbstractArray),
+    (:AbsVecOrMat, :NamedDimsArray),
+    (:AbstractArray, :NamedDimsArray),
+    (:NamedDimsArray, :NamedDimsArray)
+    ]
+
+    for (fun, d) in zip((:vcat, :hcat), (1, 2))
+
```
oxinabox, line 209:
```suggestion
```

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+        @eval function Base.$fun(a::$T, b::$S)
+            return Base.cat(a, b, dims=$d)
+        end
+
+        @eval function Base.$fun(a::$T, b::$S, c::NamedDimsArray...)
+            return Base.cat(a, b, c..., dims=$d)
+        end
+
```
oxinabox, line 217:
```suggestion=
```

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    ]
+
+    for (fun, d) in zip((:vcat, :hcat), (1, 2))
+
+        @eval function Base.$fun(a::$T, b::$S)
+            return Base.cat(a, b, dims=$d)
+        end
+
+        @eval function Base.$fun(a::$T, b::$S, c::NamedDimsArray...)
+            return Base.cat(a, b, c..., dims=$d)
```
oxinabox, line 215:
```suggestion
            return Base.cat(a, b, c...; dims=$d)
```

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    (:NamedDimsArray, :AbstractArray),
+    (:AbsVecOrMat, :NamedDimsArray),
+    (:AbstractArray, :NamedDimsArray),
+    (:NamedDimsArray, :NamedDimsArray)
+    ]
+
+    for (fun, d) in zip((:vcat, :hcat), (1, 2))
+
+        @eval function Base.$fun(a::$T, b::$S)
+            return Base.cat(a, b, dims=$d)
```
oxinabox, line 211: per the style-guide:
https://github.com/invenia/BlueStyle#keyword-arguments

```suggestion
            return Base.cat(a, b; dims=$d)
```

test/functions.jl
```diff
@@ -28,6 +28,85 @@ using Statistics
...
+
+            @testset "dimension requirements" begin
+                @test_throws DimensionMismatch f(nda, nda')
+                @test_throws DimensionMismatch f(nda, nda, nda')
+                
+                @test dimnames(f(nda, nda)) == dimnames(cat(nda, nda, dims=d))
+                @test dimnames(f(nda, a)) == dimnames(cat(nda, a, dims=d))
+                @test dimnames(f(a, nda)) == dimnames(cat(a, nda, dims=d))
+            end
+
```
oxinabox, line 105:
```suggestion
```

test/functions.jl
```diff
@@ -28,6 +28,85 @@ using Statistics
...
+        end
+    end
```
oxinabox, line 107: these tests are more complicated and confusing than I would like.
I wonder if there is a way to write them that uses more literals so it is clearer what is happening?
There may not be

test/name_core.jl
```diff
@@ -94,6 +95,30 @@ end
     @test 0 == @ballocated (()->names_are_unifiable((:a, :b), (:a, :b)))()
 end
 
+@testset "expand_dimnames" begin
+
+    @testset "single symbol" begin
+        @test expand_dimnames((), :x) == (:x, )
+        @test expand_dimnames((), :x) == (:x, )
```
oxinabox, line 102:
```suggestion
        @test expand_dimnames((), :x) == (:x,)
        @test expand_dimnames(:x, ()) == (:x,)
```

test/name_core.jl
```diff
@@ -94,6 +95,30 @@ end
     @test 0 == @ballocated (()->names_are_unifiable((:a, :b), (:a, :b)))()
 end
 
+@testset "expand_dimnames" begin
+
```
oxinabox, line 99:
```suggestion
```

test/name_core.jl
```diff
@@ -94,6 +95,30 @@ end
...
+        @test expand_dimnames((:x, :y), (:z, :w)) == (:x, :y, :z, :w)
+        @test expand_dimnames((:x, :y), (:x, :w)) == (:x, :y, :w)
+    end
+
+    @testset "number" begin
+        @test expand_dimnames((:x, :y), 1) == (:x, :y)
+        @test expand_dimnames((:x, :y), 3) == (:x, :y)
+        @test expand_dimnames((:x, :y), (1, 2, 3)) == (:x, :y)
+    end
+
```
oxinabox, line 119:
```suggestion
```

test/functions.jl
```diff
@@ -28,6 +28,85 @@ using Statistics
...
+        end
+
+        @testset "dims argument is named" begin
+
+            @test cat(nda, nda, dims=1) == cat(nda, nda, dims=:x)
+            @test cat(nda, nda, dims=2) == cat(nda, nda, dims=:y)
+
+            @test dimnames(cat(nda, nda, dims=:z)) == (:x, :y, :z)
+        end
+
```
oxinabox, line 79:
```suggestion
```

test/functions.jl
```diff
@@ -28,6 +28,85 @@ using Statistics
...
+            @test dimnames(cat(nda, a, dims=3)) == (dimnames(nda)..., :_)
+            @test dimnames(cat(a, nda, dims=3)) == (dimnames(nda)..., :_)
+        end
+
+        @testset "dims argument is named" begin
+
+            @test cat(nda, nda, dims=1) == cat(nda, nda, dims=:x)
+            @test cat(nda, nda, dims=2) == cat(nda, nda, dims=:y)
+
+            @test dimnames(cat(nda, nda, dims=:z)) == (:x, :y, :z)
```
oxinabox, line 77:
```suggestion
            @test cat(nda, nda; dims=1) == cat(nda, nda; dims=:x)
            @test cat(nda, nda; dims=2) == cat(nda, nda; dims=:y)

            @test dimnames(cat(nda, nda; dims=:z)) == (:x, :y, :z)
```

test/functions.jl
```diff
@@ -28,6 +28,85 @@ using Statistics
...
+        end
+
+    end
+
+    for (f, d) in zip((vcat, hcat), (1, 2))
+
+        @testset "$f" begin
+
+            @testset "basic functionality" begin
+                @test f(nda, nda) == cat(nda, nda, dims=d)
```
oxinabox, line 87: Semi-colons are missing all through this bit.
and probably elsewhere
https://github.com/invenia/BlueStyle#keyword-arguments

src/functions.jl
```diff
@@ -146,6 +146,78 @@ function Base.append!(A::NamedDimsArray{L,T,1}, B::AbstractVector) where {L,T}
...
+    data = Base.cat(parent(a); dims=numerical_dims)
+    newL = unify_names_longest(newL, dimnames(data))
+    return NamedDimsArray{newL}(data)
+end
+
+function Base.cat(a::NamedDimsArray{L}, b::AbstractArray; dims) where L
+    newL = expand_dimnames(L, dims)
+    numerical_dims = dim(newL, dims)
+    data = Base.cat(parent(a), b; dims=numerical_dims)
+    newL = unify_names_longest(newL, dimnames(data)) # when dims=3 for two 2d arrays
```
oxinabox, line 164: It might be good to check closely if `data` is type-infering right and that calling `dimnames` on data returns a constant result.
We should have basically all the info we need at compile time to workout what the final names will be.
So this should be fully type stable, but it seems not to be to me, so we might need a different way to find the names than usign `dimnames` on `cat`

src/name_core.jl
```diff
@@ -71,6 +71,30 @@ Base.@pure function dim_noerror(dimnames::Tuple{Vararg{Symbol, N}}, name::Symbol
...
+    if name in dimnames
+        return dimnames
+    else
+        return (dimnames..., name)
+    end
+end
+
+function expand_dimnames(dimnames::Tuple, name::Union{Integer, Colon, Tuple{}})
+    return dimnames
+end
```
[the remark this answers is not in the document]
mzgubic: Good catch. I wasn't sure whether `expand_dimnames` is the way to go, so implemented the simplest possible approach (following YAGNI). I agree the current approach is a bit unintuitive.

How should `expand_dimnames` behave when called with an integer argument?

I guess it could be expanded with a `:_` if the number is higher than the existing number of dimensions. I.e. something like
```
julia> NamedDims.expand_dimnames((:x, :y), 1)
(:x, :y)

julia> NamedDims.expand_dimnames((:x, :y), 3)
(:x, :y, :_)
```

src/name_core.jl
```diff
@@ -71,6 +71,30 @@ Base.@pure function dim_noerror(dimnames::Tuple{Vararg{Symbol, N}}, name::Symbol
...
+and `name` being a name.
+This expands the `dimnames` if `name` is not in `dimnames`.
+e.g. `expand_dimnames((:a, :b), :c) == (:a, :b, :c)`
+If `name` is already in `dimnames` then `dimnames` is returned.
+"""
+function expand_dimnames(dimnames::Tuple, name::Symbol)
+    if name in dimnames
+        return dimnames
+    else
+        return (dimnames..., name)
```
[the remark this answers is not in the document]
mzgubic: I don't understand the comment, i.e. what a code constant is and what does it mean for it to fold (hold?). Could you please explain or provide a reference? Sorry for rookie Q.

src/name_core.jl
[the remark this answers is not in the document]
oxinabox: that sounds right to me, yes.

src/name_core.jl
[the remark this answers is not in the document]
oxinabox: I'll let wikipedia explain
https://en.wikipedia.org/wiki/Constant_folding

Basically we want to know what code is generated for constant.
Which can be be checked using the following pattern:
```julia
julia> f(names) = (names..., :zzz)
f (generic function with 1 method)

julia> @code_typed (()->f((:x,:y)))()
CodeInfo(
1 ─ %1 = Core.tuple(:x, :y, :zzz)::Tuple{Symbol,Symbol,Symbol}
└──      return %1
) => Tuple{Symbol,Symbol,Symbol}
```
see that a literal has ended up in the output.

in contrast:
```julia
julia> g(names) = (names..., Symbol("zzz" * ""))
g (generic function with 1 method)

julia> @code_typed (()->g((:x,:y)))()
CodeInfo(
1 ─ %1 = invoke Base.string("zzz"::String, ""::String)::String
│   %2 = $(Expr(:foreigncall, :(:jl_string_ptr), Ptr{UInt8}, svec(Any), 0, :(:ccall), :(%1)))::Ptr{UInt8}
│   %3 = Core.sizeof(%1)::Int64
│   %4 = $(Expr(:foreigncall, :(:jl_symbol_n), Ref{Symbol}, svec(Ptr{UInt8}, Int64), 0, :(:ccall), :(%2), :(%3), :(%3), :(%2)))::Symbol
│   %5 = Core.tuple(:x, :y, %4)::Tuple{Symbol,Symbol,Symbol}
└──      return %5
) => Tuple{Symbol,Symbol,Symbol}
```

Something that fully constant folds will tend to be super-fast (techically it should be zero but nothing will say that) and it would not allocate at all.
Which is why we often use `@ballocated` to check for this.
```julia
julia> @btime (()->f((:x,:y)))()
  1.489 ns (0 allocations: 0 bytes)
(:x, :y, :zzz)

julia> @btime (()->g((:x,:y)))()
  107.931 ns (1 allocation: 32 bytes)
(:x, :y, :zzz)
```

src/cat.jl
```diff
@@ -0,0 +1,69 @@
...
+    (:NamedDimsArray, :NamedDimsArray)
+    ]
+
+    for (fun, d) in zip((:vcat, :hcat), (1, 2))
+        @eval function Base.$fun(a::$T, b::$S)
+            return Base.cat(a, b; dims=$d)
+        end
+
+        @eval function Base.$fun(a::$T, b::$S, c::NamedDimsArray...)
+            return Base.cat(a, b, c...; dims=$d)
```
mcabbott, line 66: I don't think `hcat` should call `cat`, despite giving the same result:
```julia
julia> @btime hcat(A,B) setup=(A=rand(2,2); B=rand(2,2));
  54.956 ns (2 allocations: 176 bytes)

julia> _hcat(xs...) = cat(xs...; dims=2);

julia> @btime _hcat(A,B) setup=(A=rand(2,2); B=rand(2,2));
  3.812 μs (30 allocations: 1.36 KiB)
```
Also I think you want `c::AbstractArray...` as above; and this method should also cover the 2-arg case.

mzgubic: Wow, that's a huge difference. I guess the alternative is to implement it in the same way as cat, i.e. calling hcat with parent arrays?

Right, I agree, `c::AbstractArray` was what I tried first but got method ambiguities. Will see if it can work this way with the alternative implementation.

I thought splatting `c` means that 2-arg case is also covered? i.e.
```
julia> function myfunc(a::Int, b::Int, c::Float64...)
           return 0
       end

julia> myfunc(3, 3, 3.0)
0

julia> myfunc(3, 3)
0

```

mcabbott: Ambiguities are a pain, I didn't check. But yes, re "splatting c means that 2-arg case is also covered" that's what I was trying to say too. 

I think things like `cat(a, cat(b, cat(c...;` will also be very slow. Maybe `unify_names_longest` needs a vararg method, like `unify_names` has (which is not slow because it's all tuples).

src/name_core.jl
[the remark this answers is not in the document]
mzgubic: Done, and checked that there is no allocation.

src/name_core.jl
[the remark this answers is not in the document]
mzgubic: Thanks for the explanation! I've managed to write the method with `name::Integer` without allocations, but the current implementation of the method where `name::Symbol` is not constant folding for all values of the `name` argument.


It works fine for when `name` is already in `dimnames`:
```
julia> @code_typed (()->NamedDims.expand_dimnames((:x, :y), :x))()
CodeInfo(
1 ─     return (:x, :y)
) => Tuple{Symbol,Symbol}
```

But runs into problems when `name` is not in `dimnames`. It seems like the issue is that it doesn't realise what the return type will be (hence the Union). I don't know how to fix this (it works automagically when `name::Integer`)
```
julia> @code_typed (()->NamedDims.expand_dimnames((:x, :y), :z))()
CodeInfo(
1 ──       goto #12 if not true
2 ┄─ %2  = φ (#1 => :x, #11 => %18)::Symbol
│    %3  = φ (#1 => 2, #11 => %19)::Int64
│    %4  = (%2 === :z)::Bool
└───       goto #4 if not %4
3 ──       goto #13
4 ── %7  = Base.sle_int(1, %3)::Bool
└───       goto #6 if not %7
5 ── %9  = Base.sle_int(%3, 2)::Bool
└───       goto #7
6 ──       nothing::Nothing
7 ┄─ %12 = φ (#5 => %9, #6 => false)::Bool
└───       goto #9 if not %12
8 ── %14 = Base.getfield((:x, :y), %3, false)::Symbol
│    %15 = Base.add_int(%3, 1)::Int64
└───       goto #10
9 ──       goto #10
10 ┄ %18 = φ (#8 => %14)::Symbol
│    %19 = φ (#8 => %15)::Int64
│    %20 = φ (#8 => false, #9 => true)::Bool
│    %21 = Base.not_int(%20)::Bool
└───       goto #12 if not %21
11 ─       goto #2
12 ┄       goto #13
13 ┄ %25 = φ (#3 => true, #12 => false)::Bool
└───       goto #15 if not %25
14 ─       goto #16
15 ─ %28 = Core.tuple(:x, :y, :z)::Tuple{Symbol,Symbol,Symbol}
└───       goto #16
16 ┄ %30 = φ (#14 => (:x, :y), #15 => %28)::Union{Tuple{Symbol,Symbol}, Tuple{Symbol,Symbol,Symbol}}
└───       return %30
) => Union{Tuple{Symbol,Symbol}, Tuple{Symbol,Symbol,Symbol}}
```

Funny enough, if I change the line from `if name in dimnames` to `if false`, then
```
julia> @code_typed (()->NamedDims.expand_dimnames((:x, :y), :z))()
CodeInfo(
1 ─ %1 = Core.tuple(:x, :y, :z)::Tuple{Symbol,Symbol,Symbol}
└──      return %1
) => Tuple{Symbol,Symbol,Symbol}
```

So it looks like the issue (comparing to the other method, where `if name <= length(dimnames)`) has to do with how
the compiler deals with the if statement when there is an iteration involved.

Any ideas on how to fix this?

src/cat.jl
mzgubic: Ah, yes, I understand your comment now - I had to add the separate 2-arg methods because of ambiguities as well.

test/functions.jl
mzgubic: I couldn't find a way to include literals, but have rearranged the tests somewhat and added some comments.

src/cat.jl
mzgubic: Okay, all done now. Had to replace the my much beloved tuple comprehension (yay Python) with `ntuple(i->dimnames(c[i]), length(c))`.

```
julia> @ballocated (()->cat(a, a, a, dims=1))()
1824
julia> @ballocated (()->cat(nda, nda, nda, dims=1))()
1824

julia> @ballocated (()->vcat(nda, nda, nda))()
176
julia> @ballocated (()->vcat(a, a, a))()
176
```

Thank you @mcabbott for flagging this!

mzgubic:
> finish off last few things then can merge

@oxinabox I've added some extra changes to solve the `hcat` and `vcat` problem, as well as calling `cat/vcat/hcat` with multiple arguments.

I've checked allocations and timing and it should be ok but since the change was not entirely trivial I thought you may have wanted to take one final look before merging.

test/name_core.jl
```diff
@@ -84,11 +88,15 @@ end
     if VERSION >= v"1.1"
         @test 0 == @ballocated (()->unify_names_longest((:a, :b), (:a, :_, :c)))()
         @test 0 == @ballocated (()->unify_names_shortest((:a, :b), (:a, :_, :c)))()
+        @test 0 == @ballocated (()->unify_names_longest((:a,), (:a, :b), (:a, :_, :c)))()
+        @test 0 == @ballocated (()->unify_names_shortest((:a,), (:a, :b), (:a, :_, :c)))()
```
oxinabox, line 92: Good test to add.
I was worried about this one since recusion and inlining sometime don't like each other, and inlining is a precondition for constant folding

oxinabox (approved):
LGTM

PR #131, opened 2020-09-13
rename with arbitrary number of pairs

### LINKED ISSUE #129: add `rename(nda, :oldname=>:newname)`
It would be handy to be able to do renaming via giving the old and new name.
this would work like `Base.replace(::AbstractString, old=>new)`

Right now the only way to rename is to give a full list of what the new names should be

mzgubic:
Closes #129

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -128,6 +128,26 @@ function permute_dimnames(dimnames::NTuple{N, Symbol}, perm) where N
 end
 
 """
+    _rename(names::Tuple{Vararg{Symbol}}, old_new::Vararg{Pair{Symbol, Symbol}})
+    _rename(name::Symbol, old_new::Vararg{Pair{Symbol, Symbol}})
+
+For each pair `old=>new`, replace all occurences of `old` in tuple `namea` with `new`.
+If `name` is a Symbol, replace it with `new` of the pair with matching `old`, or
+return `name` if no pairs match.
+"""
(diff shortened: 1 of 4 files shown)


src/name_core.jl
```diff
@@ -127,6 +127,26 @@ function permute_dimnames(dimnames::NTuple{N, Symbol}, perm) where N
...
+"""
+function _rename(dimnames::Tuple, old_new::Vararg{Pair})
+    return ntuple(i -> _rename(dimnames[i], old_new...), length(dimnames))
+end
+
+function _rename(dimname::Symbol, old_new::Vararg{Pair})
+    # Avoid looping over pairs explicitly because that allocates.
+    nt = ntuple(i -> dimname === first(old_new[i]) ? i : 0, length(old_new))
+    which = sum(nt)
+    which > length(old_new) && throw(ArgumentError("`old_new` needs unique `old` items"))
```
oxinabox, line 146: Can this error message be made clearer?
Maybe `duplicate old names are not permitted in rename` or something?>

oxinabox:
CI is failing

mzgubic:
> CI is failing

Yes, see: #128 and #112

src/name_core.jl
mzgubic: thanks, that's much better

mzgubic:
@oxinabox CI passes now thanks for the `test_modern` macro

oxinabox:
Tests on AV  fail because they are testing 1.1 which we don't actuall care about. 
See #133

oxinabox (approved):
Merge and tag when ready

PR #133, opened 2020-09-15
Only test LTS and stable

oxinabox:
Tests are actually failing on 1.1 right now, we only really are interested in supporting  testing LTS and Stable.
The failures are pretty meaningless


oxinabox:
Possibly we should just stop testing with AppVeyor.

mzgubic (approved):

PR #132, opened 2020-09-14
fix ambiguity on julia 1.5

oxinabox:
Closes #112
Makes #128 redundant

Cause is:
https://github.com/JuliaLang/julia/pull/36679
 removed a defintiion of matmul with `Adjoint{<:Any}` and replaced it with one with `Adjoint{<:Number}`
we were winning that ambiguity fight before, but now it ties.
This PR changes us to match.

PR #134, opened 2020-09-16
`hcat` tricky cases

mcabbott:
I was chasing ambiguities, and found the following bugs in the handling of `hcat` here: 
```julia
julia> using NamedDims

julia> nda = NamedDimsArray(rand(2,3), (:x, :y));

julia> ndv = NamedDimsArray(rand(2), (:x,));

julia> hcat(nda, ndv) # has lost a name
2×4 NamedDimsArray{(:x, :_), Float64, 2, Matrix{Float64}}:
 0.434524    0.73716   0.509066  0.309907
 0.00329738  0.669837  0.154734  0.405725

julia> hcat(nda, nda.data') # wrong sizes, but no error
2×6 NamedDimsArray{(:x, :y), Float64, 2, Matrix{Float64}}:
 0.434524    0.73716   0.509066  0.434524    0.73716   0.509066
 0.00329738  0.669837  0.154734  0.00329738  0.669837  0.154734
```
So I tried to fix them. Existing tests are untouched, and all pass. On the new testset, NamedDims v0.2.26 gives 2 passed, 4 failed, 3 errored, 0 broken.

I also replaced some `Array{T,N}` with `typeof(data)`, although I didn't make a test... I guess one could add SparseArrays or something. (Actually, I'm not sure there's any benefit to writing more than `NamedDimsArray{newL}(data)` each time.)

Edit -- this also broke `hcat` on 3-arrays, at first. Base routes these to `cat(As...; dims=Val(2))` so I made that work.

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -5,24 +5,6 @@ function Base.cat(a::NamedDimsArray{L}; dims) where L
     return NamedDimsArray{newL}(data)
 end
 
-# While the following two functions are covered by the general case below where splatted c
-# is empty, they are included because they reduce allocations to that of regular arrays.
-function Base.cat(a::NamedDimsArray{L}, b::AbstractArray; dims) where L
-    newL = expand_dimnames(L, dims)
-    numerical_dims = dim(newL, dims)
-    data = Base.cat(parent(a), b; dims=numerical_dims)
-    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
-end
-
-function Base.cat(a::AbstractArray, b::NamedDimsArray{L}; dims) where L
-    newL = expand_dimnames(L, dims)
-    numerical_dims = dim(newL, dims)
-    data = Base.cat(a, parent(b); dims=numerical_dims)
-    T = promote_type(eltype(a), eltype(b))
-    N = length(newL)
-    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
-end
-
 # to dispatch on the first _or the second_ argument being the NDA.
 for (T, S) in [
     (:NamedDimsArray, :AbstractArray),
@@ -62,4 +44,3 @@ for (T, S) in [
         end
     end
 end
-
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -2,9 +2,7 @@ function Base.cat(a::NamedDimsArray{L}; dims) where L
     newL = expand_dimnames(L, dims)
     numerical_dims = dim(newL, dims)
     data = Base.cat(parent(a); dims=numerical_dims) # Base.cat is type unstable
-    T = eltype(a)  # therefore the element type has to be inferred manually
-    N = length(newL)  # as must the size of the array
-    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
+    return NamedDimsArray{newL}(data)
 end
 
 # While the following two functions are covered by the general case below where splatted c
@@ -37,14 +35,14 @@ for (T, S) in [
         newL = expand_dimnames(combL, dims)
         numerical_dims = dim(newL, dims)
         data = Base.cat(unname(a), unname(b), unname.(cs)...; dims=numerical_dims)
-        return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
+        return NamedDimsArray{newL}(data)
     end
 end
 
 function Base.hcat(a::NamedDimsArray{L}) where L
     newL = expand_dimnames(L, 2)
     data = Base.hcat(parent(a))
-    return NamedDimsArray{newL, eltype(a), ndims(data), typeof(data)}(data)
+    return NamedDimsArray{newL}(data)
 end
 
 Base.vcat(a::NamedDimsArray{L}) where L = a
@@ -60,7 +58,7 @@ for (T, S) in [
             combL = unify_names_longest(dimnames(a), dimnames(b), dimnames.(cs)...)
             newL = expand_dimnames(combL, $d)
             data = Base.$fun(unname(a), unname(b), unname.(cs)...)
-            return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
+            return NamedDimsArray{newL}(data)
         end
     end
 end
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -4,7 +4,7 @@ function Base.cat(a::NamedDimsArray{L}; dims) where L
     data = Base.cat(parent(a); dims=numerical_dims) # Base.cat is type unstable
     T = eltype(a)  # therefore the element type has to be inferred manually
     N = length(newL)  # as must the size of the array
-    return NamedDimsArray{newL, T, N, Array{T,N}}(data)
+    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
 end
 
 # While the following two functions are covered by the general case below where splatted c
@@ -13,9 +13,7 @@ function Base.cat(a::NamedDimsArray{L}, b::AbstractArray; dims) where L
     newL = expand_dimnames(L, dims)
     numerical_dims = dim(newL, dims)
     data = Base.cat(parent(a), b; dims=numerical_dims)
-    T = promote_type(eltype(a), eltype(b))
-    N = length(newL)
-    return NamedDimsArray{newL, T, N, Array{T,N}}(data)
+    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
 end
 
 function Base.cat(a::AbstractArray, b::NamedDimsArray{L}; dims) where L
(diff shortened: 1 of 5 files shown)


src/wrapper_array.jl
```diff
@@ -60,6 +60,7 @@ Base.parent(x::NamedDimsArray) = getfield(x, :data)
 
 dim(a::NamedDimsArray{L}, name) where L = dim(L, name)
 
+NamedDimsVecOrMat{L,T} = Union{NamedDimsArray{L,T,1}, NamedDimsArray{L,T,2}}
```
mzgubic, line 63:
```suggestion
```

test/cat.jl
```diff
@@ -94,4 +94,36 @@ for (f, d) in zip((vcat, hcat), (1, 2))
             @test dimnames(f(a, nda)) == dimnames(cat(a, nda; dims=d))
         end
     end
+    @testset "more than 2 dimensions" begin
```
mzgubic, line 97: Could you please put these two `@testset`s outside the loop over `hcat` and `vcat`?

test/cat.jl
```diff
@@ -94,4 +94,36 @@ for (f, d) in zip((vcat, hcat), (1, 2))
...
+
+        @test_throws Exception hcat(nda, r23')
+        @test_throws Exception cat(nda, r23', dims=2)
+
+        @test size(cat(nda', r23', nda', dims=(:x, :y))) == (9,6)
+        @test dimnames(cat(nda', r23', nda', dims=(1,2))) == (:y, :x)
+
+        @test dimnames(cat(1:2, ndv, [5 6]', dims=:z)) == (:x, :_, :z)
+
+        vcat(pi, ndv) # does not at present have names
```
mzgubic, line 126: Nice one, I'll open an issue for this

src/cat.jl
```diff
@@ -34,56 +32,36 @@ for (T, S) in [
...
-const AbsVecOrMat = Union{AbstractVector, AbstractMatrix}
 for (T, S) in [
-    (:NamedDimsArray, :AbsVecOrMat),
-    (:NamedDimsArray, :AbstractArray),
-    (:AbsVecOrMat, :NamedDimsArray),
-    (:AbstractArray, :NamedDimsArray),
-    (:NamedDimsArray, :NamedDimsArray)
+    (:NamedDimsVecOrMat, :NamedDimsVecOrMat),
+    (:NamedDimsVecOrMat, :AbstractVecOrMat),
+    (:AbstractVecOrMat, :NamedDimsVecOrMat),
```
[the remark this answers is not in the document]
mcabbott: Restricting them is the path out of dispatch hell :) Base only has special methods for 2D case, after that it calls `cat`:
```
@less hcat(ones(2,2,2), ones(2,2,2))
```

test/cat.jl
mcabbott: Oh sorry, I messed that up, will fix.

test/cat.jl
mcabbott: I messed around a bit... but not totally sure it has to be supported! 
```
function Base.vcat(xs::Union{Number, NamedDimsVector}...)
    nonames = map(xs) do x
        x isa AbstractVector ? unname(x) : x
    end
    data = vcat(nonames...)
    allL = map(xs) do x
        x isa AbstractVector ? dimnames(x) : (:_,)
    end
    newL = unify_names(allL...)
    NamedDimsArray{newL}(data)
end
```

mzgubic: I've added https://github.com/invenia/NamedDims.jl/issues/135, might be a good in for a new contributor

src/cat.jl
```diff
@@ -4,7 +4,7 @@ function Base.cat(a::NamedDimsArray{L}; dims) where L
     data = Base.cat(parent(a); dims=numerical_dims) # Base.cat is type unstable
     T = eltype(a)  # therefore the element type has to be inferred manually
     N = length(newL)  # as must the size of the array
```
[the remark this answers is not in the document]
mcabbott: I missed those. `cat` is indeed unstable, but seems no worse after deleting these lines.

src/cat.jl
```diff
@@ -4,7 +4,7 @@ function Base.cat(a::NamedDimsArray{L}; dims) where L
...
-    return NamedDimsArray{newL, T, N, Array{T,N}}(data)
+    return NamedDimsArray{newL, eltype(data), ndims(data), typeof(data)}(data)
 end
 
 # While the following two functions are covered by the general case below where splatted c
```
mcabbott, line 10: Is this still true btw? This is what I get with or without these cases:
```
julia> nda = NamedDimsArray(rand(2,3), (:x, :y)); r23 = rand(2,3);

julia> @btime cat($nda, $r23, dims=:z);
  3.988 μs (33 allocations: 1.52 KiB)
```

src/cat.jl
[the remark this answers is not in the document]
mzgubic: sounds good

src/cat.jl
mzgubic: Right I was scratching my head about all the workarounds I did to make it not allocate that were just removed here without any effect. Potentially a result of the recent changes here https://github.com/invenia/NamedDims.jl/pull/132/files

src/cat.jl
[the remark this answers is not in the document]
mzgubic: Yeah I tested as well before suggesting to delete. In fact now I am thinking we could just go back to 
`return NamedDimsArray{newL}(data)` as you suggest in the PR.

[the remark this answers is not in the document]
mcabbott: Done.

src/cat.jl
mcabbott: Ah could be, those look relevant. Shall I kill these?

src/cat.jl
```diff
@@ -34,56 +32,36 @@ for (T, S) in [
...
-    T = eltype(a)
-    N = length(newL)
-    return NamedDimsArray{newL, T, N, Array{T,N}}(data)
+    return NamedDimsArray{newL, eltype(a), ndims(data), typeof(data)}(data)
 end
 
 Base.vcat(a::NamedDimsArray{L}) where L = a
 
-# Base.hcat and Base.vcat specialise on this Union
-const AbsVecOrMat = Union{AbstractVector, AbstractMatrix}
```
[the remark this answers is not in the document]
mcabbott: There is a whole zoo, up to things like `AdjOrTransUpperOrUnitUpperTriangular` :)

src/cat.jl
mzgubic: with fire

mcabbott: Done!

mzgubic (approved):

src/cat.jl
```diff
@@ -34,56 +12,57 @@ for (T, S) in [
...
+    @eval begin
+        Base.reduce(::typeof($f), A::AbstractVector{<:NamedDimsVecOrMat}) =
+            $nf(mapreduce(dimnames, unify_names_longest, A), A)
+        Base.reduce(::typeof($f), A::NamedDimsVector{<:Any,<:AbstractVecOrMat}) =
+            $nf(mapreduce(dimnames, unify_names_longest, A), A)
+        Base.reduce(::typeof($f), A::NamedDimsVector{<:Any,<:NamedDimsVecOrMat}) =
+            $nf(mapreduce(dimnames, unify_names_longest, A), A)
+
+        function $nf(Linner, A)
+            Louter = ($tup..., dimnames(A)...)
```
mzgubic, line 61: Why does `hcat` get an extra unnamed dimension?

src/cat.jl
```diff
@@ -34,56 +12,57 @@ for (T, S) in [
...
-            T = promote_type(eltype(a), eltype(b), ntuple(i->eltype(c[i]), length(c))...)
-            N = length(newL)
-            return NamedDimsArray{newL, T, N, Array{T,N}}(data)
+for (f, nf, tf, tup) in [
+    (:vcat, :_named_vcat, :_typed_vcat, ()),
+    (:hcat, :_named_hcat, :_typed_hcat, (:_,)),
+    ]
+    @eval begin
+        Base.reduce(::typeof($f), A::AbstractVector{<:NamedDimsVecOrMat}) =
+            $nf(mapreduce(dimnames, unify_names_longest, A), A)
```
mzgubic, line 54: Could we just call `hcat` on `A...`? Or do things complicate in a subtle way?

src/cat.jl
mcabbott: Because the reduction is always along 2nd dimension, so the outer container’s name, if any, must not be first.

src/cat.jl
mcabbott: It’s much slower to splat, if there are many slices.

src/cat.jl
mzgubic: ok, lgtm

mcabbott: Cool. I don't actually have a working merge button on this package, but maybe you do?

PR #137, opened 2020-09-21
WIP: Run JuliaFormatter to inspect changes

nickrobinson251:
Run JuliaFormatter as a way to test out https://github.com/domluna/JuliaFormatter.jl/pull/295

Ran
```julia
julia> format(
                  "src/"; 
                  style=BlueStyle(),  # requires JuliaFormatter v0.8.2 https://github.com/domluna/JuliaFormatter.jl.git#bluestyle-1
                  always_use_return=true,
                  short_to_long_function_def=true,
                  whitespace_ops_in_indices=true,
                  remove_extra_newlines=true,
                  always_for_in=true,
                  import_to_using=true,
                  pipe_to_function_call=true,
                  whitespace_in_kwargs=false,
                  whitespace_typedefs=true,
         )
```

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -30,9 +30,7 @@ for (mod, funs) in (
 end
 
 # 1 Arg - no default for `dims` keyword
-for (mod, funs) in (
-    (:Base, (:cumsum, :cumprod, :sort, :sort!, :sortslices)),
-)
+for (mod, funs) in ((:Base, (:cumsum, :cumprod, :sort, :sort!, :sortslices)),)
     for fun in funs
         @eval function $mod.$fun(a::NamedDimsArray; dims, kwargs...)
             numerical_dims = dim(a, dims)
@@ -49,7 +47,7 @@ for (mod, funs) in (
 end
 
 if VERSION > v"1.1-"
-    function Base.eachslice(a::NamedDimsArray{L}; dims, kwargs...) where L
+    function Base.eachslice(a::NamedDimsArray{L}; dims, kwargs...) where {L}
         numerical_dims = dim(a, dims)
         slices = eachslice(parent(a); dims=numerical_dims, kwargs...)
         return Base.Generator(slices) do slice
@@ -62,9 +60,7 @@ if VERSION > v"1.1-"
 end
 
 # 1 arg before - no default for `dims` keyword
-for (mod, funs) in (
-    (:Base, (:mapslices,)),
-)
+for (mod, funs) in ((:Base, (:mapslices,)),)
     for fun in funs
         @eval function $mod.$fun(f, a::NamedDimsArray; dims, kwargs...)
             numerical_dims = dim(a, dims)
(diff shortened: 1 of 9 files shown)


src/name_core.jl
```diff
@@ -288,7 +281,8 @@ function unify_names_longest(names_a, names_b)
     # 0 Allocations: @btime (()-> unify_names_longest((:a,:b), (:a,)))()
 
     length(names_a) === length(names_b) && return unify_names(names_a, names_b)
-    long, short = length(names_a) > length(names_b) ? (names_a, names_b) : (names_b, names_a)
+    long, short =
+        length(names_a) > length(names_b) ? (names_a, names_b) : (names_b, names_a)
```
nickrobinson251, line 285: I think it's that this should become
```julia
long, short = if length(names_a) > length(names_b)
    (names_a, names_b)
else
   (names_b, names_a)
end
```
and it's on the to-do list https://github.com/domluna/JuliaFormatter.jl/issues/283#issue-694066362

So it's fine that the formatter did this for now -- it's one of the not-yet-implemented things.

nickrobinson251 (commented):
Formatter comments:
- most of the diff is due to changes in the typedef whitespace e.g. `{L,T}` -> `{L, T}`
    - BlueStyle doesn't have an opinion on this, and this pacakge is inconsistent in it's use of space here
    - I set `whitespace_typedefs=true` as more often then not this package uses whitespace (and i wanted to minimise diffs so as to see the other changes), but `whitespace_typedefs=false` would also be an acceptable choice (and also make a lot of changes)
   - BlueStyle needs to pick a preference here (at least, the `BlueStyle()` in JuliaFormatter needs to be told what to do by default)
- Many changes are due to `where S` -> `where {S}`
  - again, i think this pacakge is currently inconsistent on this
  - and again, i don't think BlueStyle has a preference on this
  - seems fine
- Bar one (not-yet-implemented) case with ternary operators that go over the line length, all other changes simply "improve" style in-keeping wiht BlueStyle

Basically this is near perfect, as far as i can see!

nickrobinson251:
> I set whitespace_typedefs=true as more often then not this package uses whitespace

I think i was wrong about this, and it's less pretty in my opinion, see #138

src/functions.jl
```diff
@@ -174,24 +166,43 @@ for (T, S) in [
...
-Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
+function Base.collect(
+    x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray, Vararg{Any}}}},
+)
+    return collect_product(x)
+end
+function Base.collect(
+    x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{
+        <:Any, <:NamedDimsArray, Vararg{Any},
+    }}},
```
oxinabox, line 195: I am not sure about this

nickrobinson251: See #138

nickrobinson251: it's a consequence of the original code ignoring line limits, and the formatter making the best attempt at a tough job (it did fine in my opinion), but i fixed it up to something nicer in #138

PR #138, opened 2020-09-21
Run formatter (without typedef whitespace)

nickrobinson251:
Same as #137 but with `whitespace_typedefs=false` (i.e. Run JuliaFormatter as a way to test out domluna/JuliaFormatter.jl#295)

```julia
julia> format(
                  "src/"; 
                  style=BlueStyle(),  # requires JuliaFormatter v0.8.2 https://github.com/domluna/JuliaFormatter.jl.git#bluestyle-1
                  always_use_return=true,
                  short_to_long_function_def=true,
                  whitespace_ops_in_indices=true,
                  remove_extra_newlines=true,
                  always_for_in=true,
                  import_to_using=true,
                  pipe_to_function_call=true,
                  whitespace_in_kwargs=false,
                  whitespace_typedefs=false,
         )
```
(fyi it's also possible to tell some bits to not be formatted https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/)

### DIFF
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -1,4 +1,4 @@
-function Base.cat(a::NamedDimsArray{L}; dims) where L
+function Base.cat(a::NamedDimsArray{L}; dims) where {L}
     newL = expand_dimnames(L, dims)
     numerical_dims = dim(newL, dims)
     data = Base.cat(parent(a); dims=numerical_dims) # Base.cat is type unstable
@@ -9,9 +9,8 @@ end
 for (T, S) in [
     (:NamedDimsArray, :AbstractArray),
     (:AbstractArray, :NamedDimsArray),
-    (:NamedDimsArray, :NamedDimsArray)
-    ]
-
+    (:NamedDimsArray, :NamedDimsArray),
+]
     @eval function Base.cat(a::$T, b::$S, cs::AbstractArray...; dims)
         combL = unify_names_longest(dimnames(a), dimnames(b), dimnames.(cs)...)
         newL = expand_dimnames(combL, dims)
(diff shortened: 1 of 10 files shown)


src/functions_math.jl
```diff
@@ -58,11 +62,11 @@ with 2D type parameterization `MatrixT` and 1D `VectorT`.
 It defines the various overloads for `Base.:*` that are required.
 It should be used at the top level of a module.
 """
-macro declare_matmul(MatrixT, VectorT=nothing)
-    dim_combos = VectorT === nothing ? ((2,2),) : ((1,2), (2,1), (2,2))
+macro declare_matmul(MatrixT; VectorT=nothing)
```
nickrobinson251, line 65: oooo, this is a bug!

nickrobinson251:
Manually fixed up a couple things (see commits) :) 

I opened this PR as a test of the Formatter, but since it works so well, i suggest we merge it 😊

src/cat.jl
```diff
@@ -45,17 +43,18 @@ for (T, S) in [
     end
 end
 
-for (f, nf, tf, tup) in [
-    (:vcat, :_named_vcat, :_typed_vcat, ()),
-    (:hcat, :_named_hcat, :_typed_hcat, (:_,)),
-    ]
+for (f, nf, tf, tup) in
+    [(:vcat, :_named_vcat, :_typed_vcat, ()), (:hcat, :_named_hcat, :_typed_hcat, (:_,))]
```
mzgubic, line 47: I think this was clearer before

mzgubic (approved):
That's really cool, I need to look into how this formatting works :)

src/cat.jl
nickrobinson251: 🤷

nickrobinson251: i agree. And this is the downside of a formatter, even though you can encode a lot of subtle rules, it's hard to write one that captures the subtley of this particular case. We could fix this one up, but it'd change again if we re-ran the formatter so i'm inclined to leave it

src/functions.jl
```diff
@@ -174,24 +166,33 @@ for (T, S) in [
...
 
-Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
-Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:Any,<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
-Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tuple{<:NamedDimsArray,<:NamedDimsArray,Vararg{Any}}}}) = collect_product(x)
+for Tup in (
+    Tuple{<:NamedDimsArray,Vararg{Any}},
+    Tuple{<:Any,<:NamedDimsArray,Vararg{Any}},
+    Tuple{<:NamedDimsArray,<:NamedDimsArray,Vararg{Any}},
+)
+    @eval function Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:Tup}})
```
nickrobinson251, line 192:
```suggestion
    @eval function Base.collect(x::Base.Generator{<:Iterators.ProductIterator{<:$Tup}})
```

mzgubic:
`whitespace_in_kwargs=false` seems reasonable, is there a way of making it the default? So that it is possible to just run `format(".", style=BlueStyle())`?

nickrobinson251:
> is there a way of making it the default? So that it is possible to just run format(".", style=BlueStyle())?

Not yet, but hopefully soon https://github.com/domluna/JuliaFormatter.jl/issues/296

PR #143, opened 2020-10-22
Add linalg factorizations

rofinn:
- Reduce required code by making a `NamedFactorization` type

### DIFF
--- a/src/functions_linearalgebra.jl
+++ b/src/functions_linearalgebra.jl
@@ -1,32 +1,35 @@
 
 #==
     Note on implementation of factorisition types:
-    The general strategy is to make one of the fields of the factorization types
-    into a NamedDimsArray.
-    We can then dispatch on this as required, and strip it out with `parent` for accessing operations.
-    However, the type of the field does not actually always match to what it should be
-    from a mathematical perspective.
-    Which is corrected in `Base.getproperty`, as required
+    The general strategy is to create a named factorization type which wraps other factorizations.
+    This allows us to define `getproperty` methods which return `NamedDimsArray`s around the
+    corresponding properties of the parent factorization types.
+
+    NOTE: Maybe we could just have a `Named` type that works with arrays and factorizations
+    kind of like `Adjoint`?
 ==#
 
-## lu
-
-function LinearAlgebra.lu!(nda::NamedDimsArray{L}, args...; kwargs...) where L
-    inner_lu = lu!(parent(nda), args...; kwargs...)
-    factors = NamedDimsArray{L}(getfield(inner_lu, :factors))
-    ipiv = getfield(inner_lu, :ipiv)
-    info = getfield(inner_lu, :info)
-    return LU(factors, ipiv, info)
+struct NamedFactorization{L, T, F<:Factorization{T}} <: Factorization{T}
+    fact::F
 end
 
-function Base.parent(fact::LU{T,<:NamedDimsArray{L}}) where {T, L}
-    factors = parent(getfield(fact, :factors))
-    ipiv = getfield(fact, :ipiv)
-    info = getfield(fact, :info)
-    return LU(factors, ipiv, info)
+# Need parent to explicitly call `getfield` to avoid infinitely calling `getproperty`
+Base.parent(named::NamedFactorization) = getfield(named, :fact)
+Base.size(named::NamedFactorization) = size(parent(named))
+
+# Convenience constructors
+for func in (:lu, :lu!, :lq, :lq!, :svd, :svd!, :qr, :qr!)
+    @eval begin
+        function LinearAlgebra.$func(nda::NamedDimsArray{L, T}, args...; kwargs...) where {L, T}
+            fact = $func(parent(nda), args...; kwargs...)
+            return NamedFactorization{L, T, typeof(fact)}(fact)
+        end
+    end
 end
 
-function Base.getproperty(fact::LU{T,<:NamedDimsArray{L}}, d::Symbol) where {T, L}
+# `getproperty` wrappers
+## LU
+function Base.getproperty(fact::NamedFactorization{L, T, <:LU}, d::Symbol) where {L, T}
     inner = getproperty(parent(fact), d)
     n1, n2 = L
     if d === :L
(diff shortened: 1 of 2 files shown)


test/functions_linearalgebra.jl
```diff
@@ -31,18 +31,20 @@ end
 @testset "svd" begin
     nda = NamedDimsArray{(:foo, :bar)}([1.0 2; 3 4])
     x = svd(nda)
+    # Test based on visualization on wikipedia
+    # https://en.wikipedia.org/wiki/File:Singular_value_decomposition_visualisation.svg
     @test dimnames(x.U) == (:foo, :_)
     @test dimnames(x.S) == (:_,)
```
mcabbott, line 37: It might be neat if `S` could also have a name. This could easily be stored `NamedFactorization{(:foo, :bar, :svd)}`, and provided... maybe as an extra keyword, `svd(M; name=:svd)`?

I would even be tempted to make the default something other than `:_`. It could be the function name, `:svd`, or it could be something it's very unlikely people would type, like `:∑`.

rofinn: What would the benefit be though? You can figure out that it's an `SVD` from the factorization type it's parameterized by. I suppose it'd be nice if you could retain the original names when you do `Diagonal(x.S)`, but that seems like a weird special case.

mcabbott: It would mean that `x.Vt * A` with some other `A` would retain this marker of where that dimension came from.

rofinn: Okay, but then in that case we'd need a different symbol than `:∑` for each case. I suppose we could insert the name of the function that would have inserted `:_` more generally, but that feels like it's solving a different problem.

src/functions_linearalgebra.jl
```diff
@@ -1,32 +1,53 @@
...
-    into a NamedDimsArray.
-    We can then dispatch on this as required, and strip it out with `parent` for accessing operations.
-    However, the type of the field does not actually always match to what it should be
-    from a mathematical perspective.
-    Which is corrected in `Base.getproperty`, as required
+    The general strategy is to create a named factorization type which wraps other factorizations.
+    This allows us to define `getproperty` methods which return `NamedDimsArray`s around the
+    corresponding properties of the parent factorization types.
+
+    NOTE: Maybe we could just have a `Named` type that works with arrays and factorizations
```
rofinn, line 8: It'd be nice if we could just have a `Named` type that could work more generally on arrays, factorizations and distributions as needed.

src/functions_linearalgebra.jl
```diff
@@ -1,32 +1,53 @@
...
 
-## lu
+struct NamedFactorization{L, T, F<:Factorization{T}} <: Factorization{T}
+    fact::F
+end
+
+# Need parent to explicitly call `getfield` to avoid infinitely calling `getproperty`
+Base.parent(named::NamedFactorization) = getfield(named, :fact)
+Base.size(named::NamedFactorization) = size(parent(named))
+Base.propertynames(named::NamedFactorization; kwargs...) = propertynames(parent(named))
```
rofinn, line 19: It'd be nice if there was a `Factorization` test suite to check if my type fulfills the `Factorization` type appropriately?

test/functions_linearalgebra.jl
oxinabox: I had this debate extensively, though perhaps not publically, and perhaps only internally in my head when writing these originally.
Its out of scope for this PR.
Weird things can happen for any names you automatically provide.
Because each factorization executed is effectively pointing at a different latent space.
On that basis one could always gensym the names since it should *only* be allowed to by multiplied by its partner created at the same time .

oxinabox: But lets not get into it in this PR.
It can be a follow up

oxinabox:
@mcabbott  I would be interested in hearing your thoughts on this approach vs #24  (which this is a PR against).
I am not sure how useful it is for the factorization to still subtype the right type.
(Which this gives up)

test/functions_linearalgebra.jl
mcabbott: Right, that's the argument agains a fixed default name other than `:_`. (Perhaps instead of gensym it could be generated from the existing names.)

But without settling that, the ability to store a name which the caller provides seems easy to do in this approach. And valuable for the same reason that attaching names to anything is valuable. I guess it's an argument in favour of this struct, vs. the `getproperty` trick which lacks anywhere to store things.

mcabbott:
I'm surprised how short `methodswith(Factorization)` and `methodswith(LU)` are. So maybe there isn't a semi-infinite quagmire of methods which need to be overloaded for this NamedFactorisation to work. 

But certainly a longer list than for the `getproperty` trick. What's the argument against doing that, just that it's ugly to store `F.U` with the wrong names internally? Or something else too?

rofinn:
The issues with the getproperty dispatch approach is that:

1. We needed to define custom constructors and parent functions for each type.
2. getproperty and getfield were inconsistent
3. This is closer to what adjoint does, AFAIK

test/functions_linearalgebra.jl
oxinabox: right, in FFT we gave ability to pass in names didn't we?
How would it look if we copies that API here?

oxinabox:
bump

rofinn:
Yeah, I'm not sure what folks wanted changed for this. Seems like naming things like `S` should be decided elsewhere? Did folks want me to reverse the wrapper type change?

oxinabox:
I think lets do this.

oxinabox (approved):

rofinn:
Okay, unless there are explicit objections I'm gonna merge this tomorrow. If we're missing specific methods I think we can probably just address those as they come up?

test/functions_linearalgebra.jl
mcabbott: For FFT I never finished the PR. But instead of `fft(A, 2)` from FFTW, it was going to allow `fft(nA, :y => :k)` to specify the new name, and for `fft(nA, :y)` generate `:y∿` (or the reverse). That's simpler, though, as the whole transformation is per-dimension.

PR #144, opened 2020-10-27
make unname accept anything, and dim accept AbstractArray + number (Close #136)

### LINKED ISSUE #136: More duck typing with `hasnames`?
AxisKeys currently defines `hasnames` for handling different types that can support named dims operations. It'd be nice if we could move that API into NamedDims.jl where I think it belongs.

https://github.com/mcabbott/AxisKeys.jl/blob/master/src/names.jl#L19

Might also help to provide a more general `NamedDims.dim` function that'll do this for you?

```
NamedDims.dim(A, d) = hasnames(A) ? NamedDims.dim(dimnames(A), d) : d
```

mcabbott:
This allows `numerical_dims = dim(A, dims)` to work for any array, by doing nothing when `A` doesn't have names. This closes #136.

It also allows `unname(A)` to accept anything, so that `map(unname, args)` can strip all dimension names without first checking which arguments are are arrays.

### DIFF
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -57,6 +57,7 @@ parent_type(::Type{<:NamedDimsArray{L,T,N,A}}) where {L,T,N,A} = A
 Base.parent(x::NamedDimsArray) = getfield(x, :data)
 
 dim(a::NamedDimsArray{L}, name) where {L} = dim(L, name)
+dim(a::AbstractArray, d) = d
 
 NamedDimsVecOrMat{L,T} = Union{NamedDimsArray{L,T,1},NamedDimsArray{L,T,2}}
 NamedDimsVector{L,T} = NamedDimsArray{L,T,1}
--- a/test/wrapper_array.jl
+++ b/test/wrapper_array.jl
@@ -1,5 +1,4 @@
 using NamedDims
-using NamedDims: names
 using SparseArrays
 using Test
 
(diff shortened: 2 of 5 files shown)


test/wrapper_array.jl
```diff
@@ -1,18 +1,24 @@
...
     end
 end
 
+@testset "function dim(array, dim)" begin
+    nda = NamedDimsArray([10 20; 30 40], (:x, :y))
+    @test dim(nda, :x) == 1
+    @test dim(nda, [:y, :x]) == [2, 1]
+    @test dim(nda, 2) == 2
+    @test dim(parent(nda), 2) == 2
+    @test_throws ArgumentError dim(nda, :nope)
```
oxinabox, line 20: I am surprised this is an `ArgumentError`, don't we normally use something more specific?

src/wrapper_array.jl
```diff
@@ -57,6 +57,7 @@ parent_type(::Type{<:NamedDimsArray{L,T,N,A}}) where {L,T,N,A} = A
 Base.parent(x::NamedDimsArray) = getfield(x, :data)
 
 dim(a::NamedDimsArray{L}, name) where {L} = dim(L, name)
+dim(a::AbstractArray, d) = d
```
oxinabox, line 60: should this error if out of bounds?

oxinabox (approved):
basically LGTM,
a few points to address, and needs a version number bump before we can merge.

src/wrapper_array.jl
mcabbott: My feeling was it's simpler to let whatever this gets fed to check, or not. Some functions like `size(A,d)` accept any positive `d`.

test/wrapper_array.jl
mcabbott: It's from here:

https://github.com/invenia/NamedDims.jl/blob/master/src/name_core.jl#L45

oxinabox: We should change that one day.
But not today

PR #148, opened 2020-11-12
Repeated names

mcabbott:
Maybe it's time to define & test the behaviour with repeated indices. This fixes:
* `sum(A, dims=:i)` to use the first dimension named `:i`, as it already did, and
* `A[i=1]` is an error if more than one dimension is named `:i`. Previously it fixed all such dimensions.

The one exception I can think of is that, for a `Diagonal` matrix, it may make sense to index both dimensions together. But maybe I have overlooked other uses? 

The indexing change did not break any tests. But may still count as a breaking change, not sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,3 @@ end
 
 You can do this to your own functions in your own packages, to add `NamedDimsArray` support.
 If you implement it for any functions in a standard library, a PR would be very appreciated.
-
-### Caveats
-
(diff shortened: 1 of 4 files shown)


oxinabox:
I know @nickrobinson251 likes for covarience matrixes that indexing with just the keyword gives you on the diagonal.
Not enough to use undefined behavour, but he thinks it is cute

PR #149, opened 2020-11-23
Fix a `vcat` ambiguity

mcabbott:
```
julia> vcat(1:3, NamedDimsArray(4:6, :x))
ERROR: MethodError: vcat(::UnitRange{Int64}, ::NamedDimsArray{(:x,), Int64, 1, UnitRange{Int64}}) is ambiguous. Candidates:
  vcat(V::AbstractVector{T}...) where T in Base at abstractarray.jl:1477
  vcat(V::AbstractVector{T} where T...) in Base at abstractarray.jl:1476
  vcat(a::AbstractVecOrMat{T} where T, b::Union{NamedDimsArray{L, T, 1, A} where A<:AbstractVector{T}, NamedDimsArray{L, T, 2, A} where A<:AbstractMatrix{T}} where T where L, cs::AbstractVecOrMat{T} where T...) in NamedDims at /Users/me/.julia/packages/NamedDims/ty09T/src/cat.jl:37
Possible fix, define
  vcat(::AbstractVector{T}, ::NamedDimsArray{L, T, 1, A} where A<:AbstractVector{T} where L, ::Vararg{AbstractVector{T}, N} where N) where T
```
So I did.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.29"
+version = "0.2.30"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/cat.jl
+++ b/src/cat.jl
@@ -43,6 +43,18 @@ for (T, S) in [
     end
 end
 
+for (T, S) in [
+    (:NamedDimsVector, :NamedDimsVector),
+    (:NamedDimsVector, :AbstractVector),
+    (:AbstractVector, :NamedDimsVector),
+]
+    @eval function Base.vcat(a::$T, b::$S, cs::AbstractVector...)
+        newL = unify_names(dimnames(a), dimnames(b), dimnames.(cs)...)
+        data = vcat(unname(a), unname(b), unname.(cs)...)
+        return NamedDimsArray{newL}(data)
+    end
+end
+
 for (f, nf, tf, tup) in
     [(:vcat, :_named_vcat, :_typed_vcat, ()), (:hcat, :_named_hcat, :_typed_hcat, (:_,))]
     @eval begin
--- a/test/cat.jl
+++ b/test/cat.jl
@@ -142,6 +142,8 @@ end
 
         @test dimnames(cat(1:2, ndv, [5 6]', dims=:z)) == (:x, :_, :z)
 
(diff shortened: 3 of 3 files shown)


mzgubic (commented):
I wonder whether we could reuse the code from line 31-44? We would have to get rid of the unions and do concrete types.

Perhaps something like this instead?
```
julia> Ts = [NamedDimsArray, AbstractVector, AbstractMatrix];

julia> for (T, S) in Iterators.product(Ts, Ts)
              println(, b)
          end
```

mcabbott:
My guess is that this is going to land you in dispatch hell, although I have not tried all variations of this idea, maybe the list can eventually be made long enough to get out. (You will need to filter out pirate methods from the list, of course.)

Notice that Base has separate methods for 1, 2, and many-dimensional cases. I think it's safest not to cut across these. The present ambiguity comes from `NamedDimsVecOrMat` being less constraining on `ndims` than `AbstractVector`, but more constraining in other ways, which is avoided by following the pattern on `ndims`.
```
[18] vcat(V::AbstractVector{T}...) where T in Base at abstractarray.jl:1477
[19] vcat(V::AbstractVector{T} where T...) in Base at abstractarray.jl:1476
[20] vcat(A::AbstractVecOrMat{T}...) where T in Base at abstractarray.jl:1540
[21] vcat(A::AbstractVecOrMat{T} where T...) in Base at abstractarray.jl:1539
[22] vcat(A::Union{LinearAlgebra.UniformScaling, AbstractVecOrMat{T} where T}...) in LinearAlgebra at /Users/me/.julia/dev/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/uniformscaling.jl:388
[23] vcat(A::AbstractArray) in Base at abstractarray.jl:1774
[24] vcat(A::AbstractArray, B::AbstractArray) in Base at abstractarray.jl:1775
[25] vcat(A::AbstractArray...) in Base at abstractarray.jl:1776
```

mzgubic (approved):

gxyd:
I'm not able to understand as why the call `vcat(1:3, NamedDimsArray(4:6, :x))` isn't dispatched to
`Base.vcat(a::AbstractVecOrMat, b::NamedDimsVecOrMat, cs::AbstractVecOrMat...)` which is already defined on https://github.com/mcabbott/NamedDims.jl/blob/41fa060ec97cd3125fe7bc567831ada01cb7fda8/src/cat.jl#L39 ?

It seems more specific than any other function definition

mcabbott:
@gxyd the method `a::AbstractVecOrMat, b::NamedDimsVecOrMat` is more specific than `V::AbstractVector{T}...` in teh 2nd argument, but less specific in the first, as it does not fix `ndims` exactly.

PR #154, opened 2021-01-19
Fix tests on 1.6

### LINKED ISSUE #152: Test failures on 1.6
PkgEval shows a crash with "signal (4): Illegal instruction", as well as some problem with `@test_modern` I think?
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/56ff37a_vs_ebcc603/NamedDims.1.6.0-beta1-73f862b30f.log

The crash is from this testset:
https://github.com/invenia/NamedDims.jl/blob/master/test/cat.jl#L64
and pasting it into the REPL seems to cause the same crash. 

But setting `f, d = hcat, 2` and then pasting the contents of the outermost loop, everything passes.

mcabbott:
Closes #152

Version is already bumped, so this needs to be merged & tagged.

### DIFF
--- a/test/test_helpers.jl
+++ b/test/test_helpers.jl
@@ -1,14 +1,14 @@
 """
     @test_modern
 
-Use this to mark tests that should pass on modern (>=v1.1) versions of julia,
+Use this to mark tests that should pass on modern (>=v1.4) versions of julia,
 but should be marked as `@test_broken` on Julia 1.0.
 For more complicated cases, or where it shouldn't be defined at all on julia versions,
 you can use `if VERSION` directly.
 """
 macro test_modern(expr...)
     if VERSION >= v"1.4"
-        return :(@test $(expr...))
+        return :(@test $(expr...)) |> esc
     else
         return nothing
     end


test/test_helpers.jl
```diff
@@ -1,14 +1,14 @@
...
-Use this to mark tests that should pass on modern (>=v1.1) versions of julia,
+Use this to mark tests that should pass on modern (>=v1.4) versions of julia,
 but should be marked as `@test_broken` on Julia 1.0.
 For more complicated cases, or where it shouldn't be defined at all on julia versions,
 you can use `if VERSION` directly.
 """
 macro test_modern(expr...)
     if VERSION >= v"1.4"
-        return :(@test $(expr...))
+        return :(@test $(expr...)) |> esc
```
mzgubic, line 11: Would
```suggestion
        return :(@test $(esc(expr...)))
```
also work? I personally find it clearer than putting everything in an `esc`

If there is a reason to have the whole thing in `esc`, should we change it to
```suggestion
        return esc(:(@test $(expr...)))
```
to be consistent with the rest of the package style?

mzgubic (approved):
Thanks for checking the package with Julia 1.6! I think it looks good, but I am not a macro expert so cum grano salis

I wonder whether we should remove the `$` from `$unify` here as well? I don't think it is necessary
https://github.com/invenia/NamedDims.jl/blob/01baaccfbedb57f26eea57fde971a227d0e40bc5/test/name_core.jl#L77

test/test_helpers.jl
mcabbott: No strong feelings, was just trying to make the minimal change to make it work. Have you checked that the suggestion does in fact work, including on 1.6? CI won't (yet) check this I don't think.

mzgubic: I've tested on the 1.6 beta that is downloadable from the website but have not built from source to include the bugfix. Both ways solve the macro problem (but the crash is still there)

mcabbott: Ok! Am just a bit concerned by that splat, since there's no 2-arg method:
```
julia> esc(1,2)
ERROR: MethodError: no method matching esc(::Int64, ::Int64)
Closest candidates are:
  esc(::Any) at essentials.jl:482
```

mzgubic: Ah great catch! This shouldn't happen (as `@test` only takes one thing too) but if it does the message is better with the current choice as `@test` would fail, not `esc`. I wonder if instead we should restrict the macro to `macro test_modern(expr)` in the first place?

mcabbott: Indeed, `@test 1 ≈ 1.1 atol=0.2` is in fact one expression, so perhaps you can just delete the `...`.

mzgubic: Is it really one expression? How about this example?
```
julia> macro mymacro(expr)
           @show expr
       end
@mymacro (macro with 1 method)

julia> @mymacro3 1≈1.1 atol=0.2
ERROR: LoadError: MethodError: no method matching @mymacro3(::LineNumberNode, ::Module, ::Expr, ::Expr)
```
The `@test` implementation actually seems to take an arbitrary number of kwargs
https://github.com/JuliaLang/julia/blob/5f10eb9085fd5a622430d020c8bc9141c7acb7c7/stdlib/Test/src/Test.jl#L372

I think we have found a reason why the `esc` should be on the outside as your originally suggest.

mcabbott: Oh I'm sorry, I tested the wrong thing:
```
julia> @pr @test 1 ≈ 1.1 atol=0.2
ex1 = :(#= REPL[312]:1 =# @test 1 ≈ 1.1 atol = 0.2)
ex2 = nothing
```
But this is a drop-in for `@test`, not a wrapper. So it does need 2 args.

mzgubic: I forget, do you have merge rights?

mcabbott: I do, but not tagging rights, I believe.

mzgubic: I don't think we need to bump a version because the PR doesn't change any code in `src`

mcabbott: I think the version is already bumped one ahead of what's tagged, because of an earlier PR. 

Would be good to tag a version which passes on 1.6- though, for the sake of the mass-testing of registered packages which gets done before release. (That's how I saw the bug in #152 in the first place.)

mzgubic: oh ok, that makes sense

PR #159, opened 2021-02-20
Drop names with a vector of linear indices

### LINKED ISSUE #158: Linear Indexing on a multidimensional array, wrong names
Pointed out in https://github.com/mcabbott/AxisKeys.jl/issues/48

```julia
julia> nda = NamedDimsArray(rand(2,3), (:vars, :obs))
2×3 NamedDimsArray(::Matrix{Float64}, (:vars, :obs)):
        → obs
↓ vars  0.193787  0.705025  0.924788
        0.350784  0.428619  0.527309

julia> nda[:]
6-element NamedDimsArray(::Vector{Float64}, (:vars,)):
↓ vars  0.19378660323367924
        0.3507837608841804
        0.7050254989780798
        0.42861915061937106
        0.9247877011245385
        0.5273088922725788

julia> nda[1:2:end]
3-element NamedDimsArray(::Vector{Float64}, (:vars,)):
↓ vars  0.19378660323367924
        0.7050254989780798
        0.9247877011245385
        
julia> nda[[1,5,6,]]
3-element NamedDimsArray(::Vector{Float64}, (:vars,)):
↓ vars  0.19378660323367924
        0.9247877011245385
        0.5273088922725788
```

Linear Indexing always returns a AbstractVector.
It's dimension doesn't corespond to either of the existing dimensions of the array being indexed into.
But we apparently just assign it the first. See all the above examples
It's dimension does corespond to the dimension of the indexing vector.

So actally the following should have the name `foo`
```julia
julia> x[NamedDimsArray{(:foo,)}([1,4,5])]
3-element Vector{Float64}:
 0.4987215341030389
 0.18635552495009877
 0.4817820740518193
 ```

### PR DESCRIPTION
**Pull request by mcabbott**
Possibly closes #158.

### DIFF
--- a/src/name_core.jl
+++ b/src/name_core.jl
@@ -381,8 +381,13 @@ remaining_dimnames_from_indexing(dn::Tuple, inds::Tuple{Vararg{<:Integer}}) = ()
 remaining_dimnames_from_indexing(dn::Tuple, ci::Tuple{CartesianIndex}) = ()
 
 function remaining_dimnames_from_indexing(
-    dimnames::Tuple{<:Any,<:Any,Vararg}, inds::Tuple{T},
-) where {T<:Union{Base.LogicalIndex,AbstractVector{<:CartesianIndex}}}
+    dimnames::Tuple{<:Any,<:Any,Vararg},  # only for ndims(A) >= 2
+    inds::Tuple{T},
+) where {T<:Union{
+    Base.LogicalIndex,  # e.g. A[A .> 0]
+    AbstractVector{<:CartesianIndex},
+    AbstractVector{<:Integer},  # e.g. A[1:end]
+}}
     return ()
 end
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by oxinabox** [comment]
LGTM, thanks
Can we add a testset for all the things that shouldn't end up with names, to make sure they don't?

E.g. `nda[:]`, `nda[1:2:end]` ?

Once we have that, bump version and merge when ready.
I think you should actually be able to tag releases now even.

---

This doens't close the thing that things hould have names that come from the  index if it is named.
But that can be a seperate PR.

```julia
julia> x = rand(2,2);

julia> ndv = NamedDimsArray{(:vec,)}([1,2]);

julia> x[ndv]
2-element Vector{Float64}:
 0.5302551843796339
 0.943562702594118
```

oxinabox (approved):

mcabbott:
> This doens't close the thing that things hould have names that come from the index if it is named.
> But that can be a seperate PR.

The reason, BTW, is that the wrapper array isn't yet defined in `name_core.jl` where this fix is applied, so I wasn't sure where to put the extra method.

mcabbott:
Oops I forgot to bump the version. Perhaps can be done in #151

PR #161, opened 2021-02-20
Pretty printing

### LINKED ISSUE #160: Add pretty-printing of NamedDimsArray from AxisKeys.jl
Noticed in #158 that `nda` had pretty printing
```julia
julia> nda = NamedDimsArray(rand(2,3), (:vars, :obs))
2×3 NamedDimsArray(::Matrix{Float64}, (:vars, :obs)):
        → obs
↓ vars  0.193787  0.705025  0.924788
        0.350784  0.428619  0.527309
```

It turns out this printing is [thanks to the excellent AxisKeys.jl](https://github.com/mcabbott/AxisKeys.jl/blob/fac28cbcd97fda791163a489f70c8dd83bbcee75/src/show.jl#L228-L246). Of course, as noted there, for AxisKeys to change the printing of NamedDims is type piracy - so this issue suggest we just move (some) of those definitions here.

There's quite a lot of printing code in AxisKeys, but [most of it is related to the "higher-dim" case](https://github.com/mcabbott/AxisKeys.jl/blob/fac28cbcd97fda791163a489f70c8dd83bbcee75/src/show.jl#L149-L226), and i think we can leave all that in AxisKeys. In my opinion, the biggest benefit comes from the addition on names on the axes, and the clearer type summary, which requires very little code. 

The code in AxixKeys is credit to @mcabbott -- would you be happy for the `print_matrix` and `summary` methods for NamedDims to be moved here?

---

For reference, here's the current printing (with NamedDims v0.2.31
```
julia> nda = NamedDimsArray(rand(2,3), (:vars, :obs))
2×3 NamedDimsArray{(:vars, :obs),Float64,2,Array{Float64,2}}:
 0.995689  0.0712981  0.335635
 0.385187  0.903711   0.970939
```
Adding the `print_matrix` method from AxisKeys:
```julia
function Base.print_matrix(io::IO, A::NamedDimsArray)
    s1 = string("↓ ", dimnames(A,1)) * "  "
    if ndims(A)==2
        s2 = string(" "^length(s1), "→ ", dimnames(A,2), "\n")
        print(io, s2)
    end
    ioc = IOContext(io, :displaysize => displaysize(io) .- (2, 0))
    Base.print_matrix(ioc, parent(A), s1)
end
```
```julia
julia> nda
2×3 NamedDimsArray{(:vars, :obs),Float64,2,Array{Float64,2}}:
        → obs
↓ vars  0.995689  0.0712981  0.335635
        0.385187  0.903711   0.970939
```
and then adding the `summary` method from AxisKeys too:
```julia
function Base.summary(io::IO, A::NamedDimsArray)
    print(io, Base.dims2string(size(A)), " NamedDimsArray(")
    Base.showarg(io, parent(A), false)
    print(io, ", ", dimnames(A), ")")
end
```
```julia
julia> nda
2×3 NamedDimsArray(::Array{Float64,2}, (:vars, :obs)):
        → obs
↓ vars  0.995689  0.0712981  0.335635
        0.385187  0.903711   0.970939
```

---

Also, while looking at the current printing, i noticed `repr` on `nda` prints like a plain Matrix:
```julia
julia> repr(nda)
"[0.9956892356266973 0.07129813156802212 0.335634613650027; 0.38518702551989015 0.9037111839600722 0.9709390165138883]"
```
So we could also add a two-arg `show` method (which wouldn't interfere with the repl pretty-printing stuff above):
```julia
function Base.show(io::IO, nda::NamedDimsArray{L}) where L
    print(io, "NamedDimsArray{", L, "}(")
    show(io, parent(nda))
    print(io, ")")
end
```
```julia
julia> repr(nda)
"NamedDimsArray{(:vars, :obs)}([0.9956892356266973 0.07129813156802212 0.335634613650027; 0.38518702551989015 0.9037111839600722 0.9709390165138883])"
```

mcabbott:
After (updated):
```
julia> ndt = NamedDimsArray(rand(Int8,30,40,1,5), (:row, :col, :_, :four));

julia> view(ndt, :, :, 1:1, 1:2)
30×40×1×2 NamedDimsArray(view(::Array{Int8, 4}, :, :, 1:1, 1:2), (:row, :col, :_, :four)):
[:, :, _=1, four=1] =
        → :col
↓ :row    14   -71    19    99   -76    75    55  …   -94    38   -85    -3   118  -108    68
          95  -113   -76   -60  -121   -37    40      -26    29   -29   -22   -91    -5  -105
          -8    22   -53    61   -56    19    -2      -50   -80   107    -5  -118     4   -56
...

julia> NamedDimsArray(OffsetArray([1 2; 3 4], -1, -1), (:x, :y))'
2×2 NamedDimsArray(adjoint(OffsetArray(::Matrix{Int64}, 0:1, 0:1)), (:y, :x)) with indices 0:1×0:1:
      → :x
↓ :y  1  3
      2  4

julia> string(ans)
"NamedDimsArray([1 3; 2 4], (:y, :x))"

julia> NamedDimsArray(sprand(3,7,0.3), (:x, :y))
3×7 NamedDimsArray(::SparseMatrixCSC{Float64, Int64}, (:x, :y)):
      → :y
↓ :x  0.0667187   ⋅    ⋅         ⋅   0.856503   ⋅         ⋅ 
       ⋅          ⋅   0.451749   ⋅    ⋅         ⋅        0.0433931
       ⋅          ⋅    ⋅         ⋅    ⋅        0.878938   ⋅ 

julia> NamedDimsArray(Diagonal(rand(Int8,4)), (:same, :same))
4×4 NamedDimsArray(::Diagonal{Int8, Vector{Int8}}, (:same, :same)):
         → :same
↓ :same  55    ⋅    ⋅    ⋅
          ⋅  -77    ⋅    ⋅
          ⋅    ⋅  -11    ⋅
          ⋅    ⋅    ⋅  110

```
Before:
```
julia> view(ndt, :, :, 1:1, 1:2)
30×40×1×2 NamedDimsArray{(:row, :col, :_, :four), Int8, 4, SubArray{Int8, 4, Array{Int8, 4}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, UnitRange{Int64}}, false}}:
[:, :, 1, 1] =
  38   -77    98   101    70   -58   -38   113  …   -11   -60    10    91    -9  106   123    -9
  62   -93    12   107    90  -126  -123   104      -33     1   -51   -52   -96  126    68    99
  57   124    63    64   -78   -69   -60   -95       82   -35     6    38    73  -52    35   -20
...

julia> NamedDimsArray(sprand(3,7,0.3), (:x, :y))
3×7 NamedDimsArray{(:x, :y), Float64, 2, SparseMatrixCSC{Float64, Int64}}:
 0.0  0.0       0.0       0.0  0.0  0.67705   0.761826
 0.0  0.0       0.0       0.0  0.0  0.0       0.0649854
 0.0  0.371099  0.972481  0.0  0.0  0.814165  0.0
```

This is a simplified version of what AxisKeys.jl does -- ~~no colour~~, and the printing of `[:, :, 1, four = 1] =` will only work on Julia 1.6.

Questions:
* whether to omit trivial dims like `[:, :, 1, four = 1] =` or to write for instance `[:, :, _=1, four=1]`.
* whether to print the `with eltype Int64` story that most Base types do. ~~Leaving this off also suppresses `with indices ...` for OffsetArrays.~~ No, in fact it does not.

Closes #160


oxinabox:
> whether to omit trivial dims like

Yes, let's leave off the `_=...`
And have it just by position

src/wrapper_array.jl
```diff
@@ -176,3 +176,43 @@ end
 @propagate_inbounds function Base.setindex!(a::NamedDimsArray, value, inds...)
     return setindex!(parent(a), value, inds...)
 end
+
+############################################
+# show
```
oxinabox, line 181: Can this be in its own file?

oxinabox (commented):
I will have a proper look at this next week

src/wrapper_array.jl
```diff
@@ -176,3 +176,43 @@ end
...
+
+# 2-arg show, mainly for repr():
+function Base.show(io::IO, A::NamedDimsArray{L,T,N}) where {L,T,N}
+    print(io, "NamedDimsArray(")
+    show(io, parent(A))
+    print(io, ", ", N == 1 ? QuoteNode(L[1]) : L, ")")
```
nickrobinson251, line 187: I think for `repr` we probably want the names in the type info, otherwise the names are separated from from the type by the arbitrarily large number of values (whereas in the `showarg`/`summary` case we're not showing the values so don't have this problem).

e.g. even in a simple case
```
julia> nda = NamedDimsArray(rand(2,3), (:x,:y));

julia> repr(nda)
"NamedDimsArray([0.7434604919275618 0.6970999852412016 0.17194407552471547; 0.01565280781692069 0.753384104139615 0.5902236599197668], (:x, :y))"
```
I think it is clearer, in this context, for it to be shown as
```
julia> nda = NamedDimsArray(rand(2,3), (:x,:y));

julia> repr(nda)
"NamedDimsArray{(:x, :y)}([0.7434604919275618 0.6970999852412016 0.17194407552471547; 0.01565280781692069 0.753384104139615 0.5902236599197668])"

File: src/wrapper_array.jl [hunk 3]
```diff
@@ -176,3 +176,43 @@ end
...
+        s2 = string(" "^length(s1), "→ :", dimnames(A,2), "\n")
+        print(io, s2)
+    end
+    ioc = IOContext(io, :displaysize => displaysize(io) .- (1, 0))
+    Base.print_matrix(ioc, parent(A), s1)
+end
+
+if VERSION > v"1.6.0-DEV.1561" # 809f27c53df7a54388a687a847e9494e0d29bd4f
+
+    function Base._show_nd_label(io::IO, A::NamedDimsArray, idxs)
```
nickrobinson251, line 209: oh nice! thanks for adding this https://github.com/JuliaLang/julia/pull/38341/files

src/wrapper_array.jl
```diff
@@ -176,3 +176,43 @@ end
...
+        print(io, "[:, :, ")
+        for i in 1:length(idxs)
+            sy = dimnames(A, i+2)
+            sy == :_ ? print(io, idxs[i]) : print(io, sy, " = ", idxs[i])
```
nickrobinson251, line 213: ah, an unfortunate thing here is that this was show "invalid" indexing. 
e.g. 
```julia
julia> nda = NamedDimsArray(rand(1, 1, 2, 2), (:a, :b, :c, :d))
1×1×2×2 NamedDimsArray(::Array{Float64, 4}, (:a, :b, :c, :d)):
[:, :, c = 1, d = 1] =
      → :b
↓ :a  0.4570253497361896

[:, :, c = 2, d = 1] =
      → :b
↓ :a  0.017185058597076752

[:, :, c = 1, d = 2] =
      → :b
↓ :a  0.9711593270392489

[:, :, c = 2, d = 2] =
      → :b
↓ :a  0.7351655482073511
```
which shows 
```julia
[:, :, c = 2, d = 2] =
      → :b
↓ :a  0.7351655482073511
```
which is not valid indexing syntax:
```julia
julia> nda[:, :, c = 2, d = 2]
ERROR: MethodError: no method matching getindex(::NamedDimsArray{(:a, :b, :c, :d), Float64, 4, Array{Float64, 4}}, ::Colon, ::Colon; c=2, d=2)
Closest candidates are:
  getindex(::NamedDimsArray, ::Any...) at /Users/npr/repos/NamedDims.jl/src/wrapper_array.jl:151 got unsupported keyword arguments "c", "d"
```

I see no reason not to show valid code whenever we can
i.e. in this case to show
```julia
[a = :, b = :, c = 2, d = 2] =
      → :b
↓ :a  0.7351655482073511
```

I also think removing the whiespace around the keyword syntax would help with readability here,
e.g.
```julia
[a=:, b=:, c=2, d=2] =
      → :b
↓ :a  0.7351655482073511
```

I can even see the case for dropping the leading colon-selected dimensions `[:, :`, and just showing
e.g.
```julia
[c=2, d=2] =
      → :b
↓ :a  0.7351655482073511
```

test/wrapper_array.jl
```diff
@@ -255,6 +255,17 @@ end
...
 
+@testset "show" begin
+    nda = NamedDimsArray([1 2; 3 4], (:x, :y))
+    @test occursin("(:x, :y)", string(nda))
+    ndv = NamedDimsArray([1,2,3], :x)
+    @test occursin("], :x)", string(ndv))
+
+    str = repr(MIME"text/plain"(), nda)
+    @test occursin("NamedDimsArray(::", str)
+    @test occursin("→ :y", str)
```
nickrobinson251, line 266: as well as having these `occursin` tests, which check certain key patterns are present, it would also be good to test some full (simple, obviously correct) cases
e.g. `@test repr(NamedDimsArray(1:3, :v)) == "..."`

test/wrapper_array.jl
```diff
@@ -255,6 +255,17 @@ end
...
+end
```
nickrobinson251, line 267: these tests are only for 2-arg `show`, right? If so, would be good to add a test or two covering the other cases :)

nickrobinson251 (commented):
thanks a lot for adding this (and so fast too)! 🚀

test/wrapper_array.jl
nickrobinson251: also, for `repr` we should test the output is parseable and evaluates back to the input (at least when `parent(nda) isa Array`s; other AbstractArrays may not have defined their own `repr` to be parseable back to its input)
i.e. `@test eval(Meta.parse(repr(nda))) == nda`

test/wrapper_array.jl
mcabbott: No, this is the 3-arg test, the `repr` form does not contain `::Matrix` (or `::Array`, depending).

src/wrapper_array.jl
mcabbott: Once you have `NamedDimsArray(rand(3,3,1,1),(:x,:_,:_,:_))` then `ans[_=1, _=1]` is still not going to be working code. I didn't think to try omitting the leading `[:, :, `.

Agree that omitting spaces looks better for one-letter names. For longer names  I'm less sure. But I do think repeatinng the names starts to look clumsy:
```
[:, :, channel = 1, year = 1] =
              → :observer
↓ :iteration  0.226816  0.268232   0.507167
              0.642586  0.898197   0.467779
              0.744658  0.0212484  0.098098

# vs.

3×3×1×1 NamedDimsArray(::Array{Float64, 4}, (:iteration, :observer, :channel, :year)):
[iteration=:, observer=:, channel=1, year=1] =
              → :observer
↓ :iteration  0.226816  0.268232   0.507167
              0.642586  0.898197   0.467779
              0.744658  0.0212484  0.098098
```

src/wrapper_array.jl
mcabbott: My thinking was that it's better to present one way of writing things, not two. I agree that `string(array)` for a large array doesn't print something very eyeball-friendly, never has. It does get trimmed in some contexts, e.g. `[NamedDimsArray(rand(Int8, 30),:x) for y in 1:4]`.

src/wrapper_array.jl
```diff
@@ -176,3 +176,43 @@ end
...
+function Base.showarg(io::IO, A::NamedDimsArray{L,T,N}, outer) where {L,T,N}
+    print(io, "NamedDimsArray(")
+    Base.showarg(io, parent(A), false)
+    print(io, ", ", N == 1 ? QuoteNode(L[1]) : L, ")")
+end
+
+function Base.print_matrix(io::IO, A::NamedDimsArray)
+    s1 = string("↓ :", dimnames(A,1)) * "  "
+    if ndims(A)==2
+        s2 = string(" "^length(s1), "→ :", dimnames(A,2), "\n")
```
mcabbott, line 200: Note that this gets the spacing wrong for some unicode, e.g. `NamedDimsArray(rand(3,4),(:x̂, :ŷ))` where `length("x̂")==2`.

src/wrapper_array.jl
mcabbott: It's also possible that this should be turned off based on IOContext... here `Int8` is not printed when the container already specifies it:
```
julia> rand(Int8, 3) |> string
"Int8[-121, 12, 63]"

julia> [rand(Int8, 3) for _ in 1:2]
2-element Vector{Vector{Int8}}:
 [19, 127, -37]
 [-110, -124, -52]

julia> Any[rand(Int8, 3) for _ in 1:2]
2-element Vector{Any}:
 Int8[13, -107, 37]
 Int8[85, 45, -111]
```

src/wrapper_array.jl
nickrobinson251: > Once you have NamedDimsArray(rand(3,3,1,1),(:x,:_,:_,:_)) then ans[_=1, _=1] is still not going to be working code.

yeah, that's true. But i don't see it as a reason not to show valid code whenever possible.

Thanks for the examples, i think the combination of omitting spaces and omitting the leading dimensions looks nicest to me
```
3×3×1×1 NamedDimsArray(::Array{Float64, 4}, (:iteration, :observer, :channel, :year)):
[channel=1, year=1] =
              → :observer
↓ :iteration  0.226816  0.268232   0.507167
              0.642586  0.898197   0.467779
              0.744658  0.0212484  0.098098
```

test/wrapper_array.jl
nickrobinson251: okay, cool - thanks!

would you mind add a comment saying which method each is testing? Julia's printing mechanism and what-calls-what is something that's quite hard to remember, so it's not always clear from the code (to me at least)

mcabbott: At first I put more of the string, to try to make it obvious, but trimmed it because of the `Array{T,2}` vs `Matrix` printing change

src/wrapper_array.jl
mcabbott: Actually, the problem is in Base, not here:
```
julia> Base.print_matrix(stdout, ones(3,4), " x: "); println()
 x: 1.0  1.0  1.0  1.0
    1.0  1.0  1.0  1.0
    1.0  1.0  1.0  1.0

julia> Base.print_matrix(stdout, ones(3,4), " x̂: "); println()
 x̂: 1.0  1.0  1.0  1.0
     1.0  1.0  1.0  1.0
     1.0  1.0  1.0  1.0
```

src/wrapper_array.jl
mcabbott: Now turned off based on whether the container has eltype `NamedDimsArray`, which is how Base handles similar things.

src/wrapper_array.jl
mcabbott: See what you think of:
```
         108   -16    48   -44  -111   -51    67      -84  -115   -79    11   -12    74  -118
         115  -111     1    73    99    43    24       74   -22  -126     6   -91  -118     0

□[_=1, four=2]
        → :col
↓ :row   -70    -7   -75   116     9    35   106  …    22   -20    93   -49     2  -111  -122
         -70   -93    70   -37   104  -126     2       88   -66    96   -48    70    41    36
         -75   -43  -126   -51    74   105   -28      -65   -38    18    29   114   -24  -115
```

AxisKeys uses `↓→□` for row/col/panel, and it looked a bit sparse without the `[:, :, ` which Base uses.

nickrobinson251: I find the little square quite confusing (and never understand what "panel" means) 

I think maybe just keeping all the names, including the leading dimensions, is maybe best.

And thanks for working on this, I think it'll be a much loved quality of life improvement

mcabbott: Not so keen to repeat, but I don't want the 3rd to be mis-read as being the first dimension, where Base puts a `:`... perhaps you should read `□` as a placeholder for the array being indexed, as a reminder that this isn't Base's printing.

Can think a bit, maybe there are other ways.

nickrobinson251: Let's see what @oxinabox thinks :)

mcabbott: Two other ideas are
```
julia> view(ndt, :, :, 1:1, 1:2)
30×40×1×2 NamedDimsArray(view(::Array{Int8, 4}, :, :, 1:1, 1:2), (:row, :col, :_, :four)):
𝒩[_=1, four=1]
        → :col
↓ :row    14   -71    19    99   -76    75    55  …   -94    38   -85    -3   118  -108    68
          95  -113   -76   -60  -121   -37    40      -26    29   -29   -22   -91    -5  -105
```
or
```
julia> view(ndt, :, :, 1, 1:2)
30×40×2 NamedDimsArray(view(::Array{Int8, 4}, :, :, 1, 1:2), (:row, :col, :four)):
(…)[four=1]
        → :col
↓ :row    14   -71    19    99   -76    75    55  …   -94    38   -85    -3   118  -108    68
          95  -113   -76   -60  -121   -37    40      -26    29   -29   -22   -91    -5  -105
          -8    22   -53    61   -56    19    -2      -50   -80   107    -5  -118     4   -56
```

oxinabox: > which shows
> 
> [:, :, c = 2, d = 2] =
>       → :b
> ↓ :a  0.7351655482073511
> 
> which is not valid indexing syntax:


Should we make that valid indexing? (if so lets open an issue about it)

I kind of think it is acceptable for show, even if we don't
All that have been shown are pretty acceptable to nme.

We can always mess with it later.
Let get this improvement in, and worry about perfection

src/wrapper_array.jl
```diff
@@ -176,3 +176,46 @@ end
...
+    Base.print_matrix(ioc, parent(A), s1)
+end
+
+if VERSION > v"1.6.0-DEV.1561" # 809f27c53df7a54388a687a847e9494e0d29bd4f
+
+    function Base._show_nd_label(io::IO, A::NamedDimsArray, idxs)
+        print(io, "□[")
+        for i in 1:length(idxs)
+            print(io, dimnames(A, i+2), "=", idxs[i])
+            i == length(idxs) ? println(io, "]") : print(io, ", ")
```
mcabbott, line 217:
```suggestion
        print(io, "[:, :, ")
        for i in 1:length(idxs)
            print(io, dimnames(A, i+2), "=", idxs[i])
            i == length(idxs) ? println(io, "] =") : print(io, ", ")
```

src/wrapper_array.jl
mcabbott: I think mixed named & positional indexing sounds super-confusing!

Have changed it back to start with `[:, :, `, this is the smallest deviation from what Base does, perhaps a good start as you say.

src/wrapper_array.jl
mcabbott: Done.

mcabbott:
I think this is ready to go, BTW. Recent commit restore colour, matching what AxisKeys prints (but with a tidier implementation).

nickrobinson251 (commented):
This looks great! ...apart from the colour - can we discuss removing the colour (or was that already discussed somewhere)?

I really do find colour harms my ability to actually read the output. and makes the repl very noisy. It seems like an accessibility nightmare in general tbh. Is there a specific argument in favour of colour? It it just seems like a fun/cute feature i'd really rather we remove it

src/wrapper_array.jl
nickrobinson251: is there an issue open to fix this in Base?

src/show.jl
```diff
@@ -0,0 +1,51 @@
...
+        show(io, parent(A))
+        print(io, ", ", N == 1 ? QuoteNode(L[1]) : L, ")")
+    end
+end
+
+# This is called by summary(), for main REPL printing:
+function Base.showarg(io::IO, A::NamedDimsArray{L,T,N}, outer) where {L,T,N}
+    print(io, "NamedDimsArray(")
+    Base.showarg(io, parent(A), false)
+    print(io, ", ", ColourString(N == 1 ? QuoteNode(L[1]) : L), ")")
```
oxinabox, line 17:
```suggestion
    print(io, ", ", ColourString(N == 1 ? QuoteNode(L[1]) : L), ")")
    return nothing
```

src/show.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+function Base.print_matrix(io::IO, A::NamedDimsArray)
+    s1 = ColourString("↓ ", dimnames(A, 1), "  ")
+    if ndims(A)==2
+        println(io, " "^Base.Unicode.textwidth(s1), ColourString("→ ", dimnames(A, 2)))
+    end
+    ioc = IOContext(io, :displaysize => displaysize(io) .- (1, 0))
+    Base.print_matrix(ioc, parent(A), s1)
```
oxinabox, line 26:
```suggestion
    Base.print_matrix(ioc, parent(A), s1)
    return nothing
```

src/show.jl
```diff
@@ -0,0 +1,51 @@
...
+# This is called by summary(), for main REPL printing:
+function Base.showarg(io::IO, A::NamedDimsArray{L,T,N}, outer) where {L,T,N}
+    print(io, "NamedDimsArray(")
+    Base.showarg(io, parent(A), false)
+    print(io, ", ", ColourString(N == 1 ? QuoteNode(L[1]) : L), ")")
+end
+
+function Base.print_matrix(io::IO, A::NamedDimsArray)
+    s1 = ColourString("↓ ", dimnames(A, 1), "  ")
+    if ndims(A)==2
```
oxinabox, line 22:
```suggestion
    if ndims(A) == 2
```

src/show.jl
```diff
@@ -0,0 +1,51 @@
...
+    ColourString(xs...) = new(string(xs...))
+end
+Base.alignment(io::IO, x::ColourString) = alignment(io, x.string)
+Base.length(x::ColourString) = length(x.string)
+Base.ncodeunits(x::ColourString) = ncodeunits(x.string)
+Base.textwidth(x::ColourString) = textwidth(x.string)
+Base.print(io::IO, x::ColourString) = printstyled(io, x.string; color=:magenta)
+
+if VERSION > v"1.6.0-DEV.1561" # 809f27c53df7a54388a687a847e9494e0d29bd4f
+
```
oxinabox, line 42:
```suggestion
```

src/show.jl
```diff
@@ -0,0 +1,51 @@
...
+    function Base._show_nd_label(io::IO, A::NamedDimsArray, idxs)
+        print(io, "[:, :, ")
+        for i in 1:length(idxs)
+            print(io, ColourString(dimnames(A, i + 2), "=", idxs[i]))
+            i == length(idxs) ? println(io, "] =") : print(io, ", ")
+        end
+    end
+
```
oxinabox, line 50:
```suggestion
```

test/wrapper_array.jl
```diff
@@ -283,6 +283,18 @@ end
     @test stride(nda, :b) == 3 == stride(nda, 2) == stride(x, 2)
 end
 
+@testset "show" begin
+    nda = NamedDimsArray([1 2; 3 4], (:x, :y))
+    @test occursin("(:x, :y)", string(nda))
+    ndv = NamedDimsArray([1,2,3], :x)
```
oxinabox, line 289:
```suggestion
    ndv = NamedDimsArray([1, 2, 3], :x)
```

oxinabox (approved):
I am inclined to say lets keep the color for now.
It is not too much color, it is just a little bit.

I personally find it increases readability.
But it is known the Nick and I are different.

We can always reconsider later and remove it.

please fix up these last few things and then merge and tag.

src/show.jl
mcabbott: I have to say, the robot and I disagree on which version is easier to read here...

PR #169, opened 2021-06-29
Use `axes` instead of `size` within `similar`

SBuercklin:
When using an `OffsetArray` for the parent array inside of a `NamedDimsArray`, calling `similar` converts the parent array to a regular dense `Array`

```
using NamedDims
using OffsetArrays

oa = OffsetArray([1,2,3], -1:1)
nda = NamedDimsArray(oa, (:x,))
parent(similar(nda)) isa OffsetArray # false
parent(similar(nda)) isa Array # true
```

This is fixed by changing `size(A)` into `axes(A)` within `named_size`. `named_size` is called to determine the length of every dimension and attach this length to its corresponding dimension name. However, when we use an `OffsetArray`, the length is not sufficient to reconstruct the array as we also need the offset. When just lengths are supplied, `OffsetArray` falls back to default `Array`s, leading to the bug.

### DIFF
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -82,25 +82,25 @@ function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
     return NamedTuple{L,NTuple{N,Int}}(size(a))
 end
-function Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T}
+function Base.similar(a::NamedDimsArray{L,T}, ::Type{eltype}=T) where {L,T,eltype}
     dnames = dimnames(a)
     return NamedDimsArray(similar(parent(a), eltype), dnames)
 end
-function Base.similar(a::NamedDimsArray, eltype::Type, dims::NamedTuple)
+function Base.similar(a::NamedDimsArray, ::Type{eltype}, dims::NamedTuple) where eltype
     new_sizes = values(dims)
     new_names = keys(dims)
     return NamedDimsArray{new_names}(similar(parent(a), eltype, new_sizes))
 end
 function Base.similar(
-    a::NamedDimsArray{L,T,N}, eltype::Type, new_names::NTuple{N,Symbol},
-) where {T,N,L}
+    a::NamedDimsArray{L,T,N}, ::Type{eltype}, new_names::NTuple{N,Symbol},
+) where {L,T,N,eltype}
     dims = NamedTuple{new_names,NTuple{N,Int}}(size(a))
     return similar(a, eltype, dims)
 end
 
 function Base.similar(
-    a::NamedDimsArray{L,T,N}, eltype::Type, new_sizes::NTuple{N,Int},
-) where {L,T,N}
+    a::NamedDimsArray{L,T,N}, ::Type{eltype}, new_sizes::NTuple{N,Int},
+) where {L,T,N,eltype}
 
     dims = NamedTuple{L,NTuple{N,Int}}(new_sizes)
     return similar(a, eltype, dims)
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -80,7 +80,8 @@ Base.axes(a::NamedDimsArray, d) = axes(parent(a), dim(a, d))
 
 function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
-    return NamedTuple{L,NTuple{N,Int}}(size(a))
+    axes_a = axes(a)
+    return NamedTuple{L,typeof(axes_a)}(axes_a)
 end
 function Base.similar(
     a::NamedDimsArray{L,T}, eltype::Type=T, dims::NamedTuple{new_names}=named_size(a),
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -80,8 +80,15 @@ Base.axes(a::NamedDimsArray, d) = axes(parent(a), dim(a, d))
 
 function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
-    axes_a = axes(a)
-    return NamedTuple{L,typeof(axes_a)}(axes_a)
+    return NamedTuple{L,NTuple{N,Int}}(size(a))
+end
+function Base.similar(a::NamedDimsArray)
+    dnames = dimnames(a)
+    return NamedDimsArray(similar(parent(a)), dnames)
+end
+function Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T}
+    dnames = dimnames(a)
+    return NamedDimsArray(similar(parent(a), eltype), dnames)
 end
 function Base.similar(
     a::NamedDimsArray{L,T}, eltype::Type=T, dims::NamedTuple{new_names}=named_size(a),
--- a/src/wrapper_array.jl
+++ b/src/wrapper_array.jl
@@ -82,22 +82,15 @@ function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
     return NamedTuple{L,NTuple{N,Int}}(size(a))
 end
-function Base.similar(a::NamedDimsArray)
-    dnames = dimnames(a)
-    return NamedDimsArray(similar(parent(a)), dnames)
-end
 function Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T}
     dnames = dimnames(a)
     return NamedDimsArray(similar(parent(a), eltype), dnames)
 end
-function Base.similar(
-    a::NamedDimsArray{L,T}, eltype::Type=T, dims::NamedTuple{new_names}=named_size(a),
-) where {L,T,new_names}
-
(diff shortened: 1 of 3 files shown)


mcabbott:
Might be worth fixing this, too, while in there:
```julia
julia> nda = NamedDimsArray(rand(10, 100), (:x, :y));

julia> similar(nda, Int, (1:2, 1:5))
2×5 OffsetArray(::NamedDimsArray{(:x, :y), Int64, 2, Matrix{Int64}}, 1:2, 1:5) with eltype Int64 with indices 1:2×1:5:
 4296441864  4296441864  4296441864  4296441864  10874420464
 4296441864  4296441864  4296441864  6205973872   4455608320

julia> similar(nda, Int, 2, 5)
2×5 NamedDimsArray{(:x, :y), Int64, 2, Matrix{Int64}}:
 4296441864  4296441864  4296441864  4296441864  10874420544
 4296441864  4296441864  4296441864  6205973872   4410804737

julia> similar(nda, Int, 3)
3-element Vector{Int64}:
 4297174192
 4297180200
 4297183464
```
I would argue that neither of these should have names, since the new dimensions don't share anything with the old -- they are preserved when `ndims` matches.

src/wrapper_array.jl
```diff
@@ -80,7 +80,8 @@ Base.axes(a::NamedDimsArray, d) = axes(parent(a), dim(a, d))
 
 function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
-    return NamedTuple{L,NTuple{N,Int}}(size(a))
+    axes_a = axes(a)
+    return NamedTuple{L,typeof(axes_a)}(axes_a)
 end
```
mcabbott, line 85: I think `named_size` ought to return the size; there could be a `named_axes` which would return this.

mcabbott: And (I can't comment on those lines) I'm not sure `similar(A)` and `similar(A, T)` should go via `named_size` at all, since it's not guaranteed to work:
```julia
julia> nda = NamedDimsArray(rand(2,2), (:x, :x));

julia> similar(nda)
ERROR: duplicate field name in NamedTuple: "x" is not unique
```
Probably they should just call `NamedDimsArray(similar(parent(A)), dimnames(A)`.

SBuercklin: This last method solves the original problem I was facing, so I'll revert `named_size` and get rid of `named_axes` entirely.

SBuercklin:
> I would argue that neither of these should have names, since the new dimensions don't share anything with the old -- they are preserved when `ndims` matches.

In this case, should we even return a `NamedDimsArray`, or should we return `similar(parent(A), ...)` as in your third example? Are there circumstances where `similar(::NamedDimsArray, ...)` should return a `NamedDimsArray` with the same names for each dimension? 

I guess the clarification I need is the following: which of the following cases should **not** return an array with named dimensions?

```julia
julia> nda = NamedDimsArray(rand(10, 100), (:x, :y));

julia> similar(nda, Int, (1:2, 1:5)) # This actually returns an `OffsetArray` wrapping a NamedDimsArray, keeps dim names
2×5 OffsetArray(::NamedDimsArray{(:x, :y), Int64, 2, Matrix{Int64}}, 1:2, 1:5) with eltype Int64 with indices 1:2×1:5:
 140352541587536  140352541587856  140352541588016  140353967309008  140352541587776
 140353967275600  140352541587936  140352541588096  140352541588176                1

julia> similar(nda, Int, (2, 5)) # Explicitly calls a constructor from NamedDimsArray, keeps dim names
2×5 NamedDimsArray{(:x, :y), Int64, 2, Matrix{Int64}}:
 0  140353937673728                1  3             -255
 1                2  140353937673729  1  140353997382352

julia> similar(nda, Int, 2, 5) # Operates the same as previous, but calls a method from Base
2×5 NamedDimsArray{(:x, :y), Int64, 2, Matrix{Int64}}:
 140353928831312  140353928836032  140353928836224  140353928839040  140353997382352
 140353928832256  140353928836112  140353928838960  140353997382352  140353997382145

julia> similar(nda, Int, 3) # Drops the dim names as provided dims do not match ndims(nda)
3-element Vector{Int64}:
 140353931628560
 140353931628624
 140353931628656
```

SBuercklin:
My concern is contradicting the existing test here (dim names propagate when you resize but don't change ndims), so I'm not sure which convention we want to consider correct: https://github.com/invenia/NamedDims.jl/blob/df644223830da175ef7650d0dbb8ec83fff50c8a/test/wrapper_array.jl#L236-L242

In either case, I pushed an update with the new `similar(nda, T)` and `similar(nda)` methods. I'm happy to do some more work on the `similar` methods as well, but I might push it to a separate PR

mcabbott (approved):
OK, sounds good. 

That looks like an explicit test of the behaviour I don't like. There may also be implicit tests, e.g. where someone relied on say `hcat` internally using `similar`, rather than explicitly handling the names here. But, later.

test/wrapper_array.jl
```diff
@@ -255,6 +256,19 @@ end
...
+        ndd = similar(ndb, Float64)
+        @test parent(ndb) !== parent(ndc)
+        @test eltype(ndc) == Float64
+        @test size(ndc) == (10, 20, 30, 40)
+        @test dimnames(ndc) == (:a, :b, :c, :d)
+        @test parent(ndc) isa typeof(oa)
+        @test parent(ndd) isa OffsetArray
+        @test eltype(parent(ndd)) == Float64
+    end
 end
```
mcabbott, line 272: I believe these changes should also fix this:
```suggestion
    end
    @testset "repeated names" begin
        ndr = NamedDimsArray([1 2; 3 4], (:same, :same))
        @test dimnames(similar(ndr)) == (:same, :same)
        @test dimnames(similar(ndr, Float64)) == (:same, :same)
    end
end
```

src/wrapper_array.jl
```diff
@@ -82,6 +82,14 @@ function named_size(a::AbstractArray{T,N}) where {T,N}
     L = dimnames(a)
     return NamedTuple{L,NTuple{N,Int}}(size(a))
 end
+function Base.similar(a::NamedDimsArray)
+    dnames = dimnames(a)
+    return NamedDimsArray(similar(parent(a)), dnames)
+end
+function Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T}
```
mcabbott, line 89: I don't think this is causing failures, but no need for default here -- that's the method just above.
```suggestion
function Base.similar(a::NamedDimsArray, eltype::Type)
```

SBuercklin: I think we need to remove the defaults from the next method as well, otherwise in both cases it will dispatch to the third one and compute the defaults. I'm working on a fix locally, I'll push it when I get it working + tested on 1.3 as well

mcabbott: There's one more quirk here, that I think most of these methods are failing to specialise on the type:
```
julia> nda = NamedDimsArray(rand(10, 100), (:x, :y));

julia> @btime similar($nda, Int);
  257.952 ns (1 allocation: 7.94 KiB)

julia> @btime similar(parent($nda), Int);
  55.287 ns (1 allocation: 7.94 KiB)

julia> Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T} = NamedDimsArray{L}(similar(parent(a), eltype))

julia> @btime similar($nda, Int);
  265.000 ns (1 allocation: 7.94 KiB)

julia> Base.similar(a::NamedDimsArray{L,T}, ::Type{eltype}=T) where {L,T,eltype} = NamedDimsArray{L}(similar(parent(a), eltype))

julia> @btime similar($nda, Int);
  50.553 ns (1 allocation: 7.94 KiB)
```
Not your fault, but it might be worth changing all to explicitly specialise?

SBuercklin: I pushed a fix changing the second argument to `::Type{eltype}`. Interestingly, I'm not able to reproduce this on my machine on 1.6.1
```julia
julia> nda = NamedDimsArray(rand(10, 100), (:x, :y));                                                                      

julia> @btime similar($nda, Int);                                                                                          
  168.334 ns (1 allocation: 7.94 KiB)

julia> @btime similar(parent($nda), Int);                                                                                  
  157.707 ns (1 allocation: 7.94 KiB)

julia> Base.similar(a::NamedDimsArray{L,T}, eltype::Type=T) where {L,T} = NamedDimsArray{L}(similar(parent(a), eltype))    

julia> @btime similar($nda, Int);                                                                                          
  158.322 ns (1 allocation: 7.94 KiB)

julia> Base.similar(a::NamedDimsArray{L,T}, ::Type{eltype}=T) where {L,T,eltype} = NamedDimsArray{L}(similar(parent(a), eltype))                                                                                                                      

julia> @btime similar($nda, Int);                                                                                          
  160.185 ns (1 allocation: 7.94 KiB)
```

mcabbott: Now that I try again, I can't get anything that dramatic, maybe 10% on one machine. I've forgotten which version the initial check was on, maybe that matters? Anyway, thanks for fixing it all the same!

PR #178, opened 2021-09-08
Add `rrule`s for constructors

### LINKED ISSUE #177: Should we add a dependency on ChainRulesCore?
I ran into 
```julia
gradient((values, names) -> sum(NamedDimsArray(values, names)), rand(2, 3), (:a, :b))

ERROR: Need an adjoint for constructor NamedDimsArray{(:node, :target), Float32, 2, Matrix{Float32}}. Gradient is of type NamedDimsArray{(:node, :target), Float32, 2, Matrix{Float32}}...
```
error when using Zygote to AD through some code using NamedDimsArrays. It can be solved by pirating
```julia
function ChainRulesCore.rrule(::Type{NamedDimsArray}, values, names)
    NamedDimsArray_pullback(ȳ::AbstractArray) = (NoTangent(), ȳ, NoTangent())
    NamedDimsArray_pullback(ȳ::Tangent) = (NoTangent(), ȳ.data, NoTangent())
    NamedDimsArray_pullback(ȳ::AbstractThunk) = NamedDimsArray_pullback(unthunk(ȳ))
    return NamedDimsArray(values, names), NamedDimsArray_pullback
end
```

The question is, do we want to add a dependency on ChainRulesCore to the package and fix this here? I think we do

mzgubic:
Closes #177

### DIFF
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -1,7 +1,11 @@
 _NamedDimsArray_pullback(ȳ::AbstractArray) = (NoTangent(), ȳ, NoTangent())
 _NamedDimsArray_pullback(ȳ::Tangent) = (NoTangent(), ȳ.data, NoTangent())
 _NamedDimsArray_pullback(ȳ::AbstractThunk) = _NamedDimsArray_pullback(unthunk(ȳ))
-
-function ChainRulesCore.rrule(::Type{NamedDimsArray}, values, names)
+function ChainRulesCore.rrule(::Type{<:NamedDimsArray}, values, names)
     return NamedDimsArray(values, names), _NamedDimsArray_pullback
+end
+
+function ChainRulesCore.rrule(T::Type{<:NamedDimsArray}, values)
+    NamedDimsArray_values_pullback(ȳ) = _NamedDimsArray_pullback(ȳ)[1:2]
+    return T(values), NamedDimsArray_values_pullback
 end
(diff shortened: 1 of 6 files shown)


oxinabox (approved):

src/chainrules.jl
```diff
@@ -0,0 +1,11 @@
+_NamedDimsArray_pullback(ȳ::AbstractArray) = (NoTangent(), ȳ, NoTangent())
```
mcabbott, line 1: I am too slow but shouldn't this remove names from the array it gets? On the assumption that similar or broadcasting in the computation of a gradient for the array with names may, probably should, produce another array with names?

```
julia> nda = NamedDimsArray(rand(Int8,3,4), (:x, :y));

julia> gradient(x -> sum(x), nda)  # does not know about names
(Fill(1, 3, 4),)

julia> gradient(x -> sum(x; dims=1)[1], nda)  # involves broadcasting, so might? 
ERROR: DimensionMismatch("destination axes (Base.OneTo(3), Base.OneTo(4)) are not compatible with source axes (Base.OneTo(1), Base.OneTo(4))")

julia> gradient(x -> sum(x; dims=1)[1], nda.data)
(Int8[1 0 0 0; 1 0 0 0; 1 0 0 0],)
```

mzgubic: thanks, I have opened an issue

PR #183, opened 2021-10-19
Add vector-vector * methods for Transpose / Adjoint

glennmoy:


src/functions_math.jl
```diff
@@ -54,10 +54,14 @@ for A in (
     Adjoint{<:Number,<:AbstractVector},
     Transpose{<:Real,<:AbstractVector{<:Real}},
     Transpose{<:Any, <:AbstractMatrix{T}} where T,  # resolves ambiguity error in AxisKeys
+    Union{Adjoint{T, <:AbstractVector{T}}, Transpose{T, <:AbstractVector{T}}} where {T},
```
mzgubic, line 57: why not just
```suggestion
    Adjoint{T, <:AbstractVector{T}} where {T},
    Transpose{T, <:AbstractVector{T}}} where {T},
```

src/functions_math.jl
```diff
@@ -54,10 +54,14 @@ for A in (
...
 )
     @eval function Base.:*(a::$A, b::NamedDimsArray{L,T,1,<:AbstractVector{T}}) where {L,T}
         return *(a, parent(b))
     end
+    @eval function Base.:*(b::NamedDimsArray{L,T,1,<:AbstractVector{T}}, a::$A) where {L,T}
```
mzgubic, line 62: Do we need this one? I think the `CoVector` one above is meant to take care of these.

mzgubic (commented):
Should we also add tests?

src/functions_math.jl
glennmoy: updated

PR #186, opened 2021-11-12
add {arg/find}{min/max} functions

### LINKED ISSUE #185: argmax and argmin
Any reason these aren't implemented?

e.g.

```
X = NamedDimsArray{(:a, :b)}(rand(2,2))
@assert argmax(X; dims=:a) == argmax(X; dims=1)
```

fredcallaway:
Fixes #185

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -17,7 +17,7 @@ end
 
 # 1 Arg
 for (mod, funs) in (
-    (:Base, (:sum, :prod, :maximum, :minimum, :extrema, :argmax, :argmin, :findmax, :findmin)),
+    (:Base, (:sum, :prod, :maximum, :minimum, :extrema, :argmax, :argmin)),
     (:Statistics, (:mean, :std, :var, :median)),
 )
     for fun in funs
@@ -25,6 +25,21 @@ for (mod, funs) in (
             numerical_dims = dim(a, dims)
             data = $mod.$fun(parent(a); dims=numerical_dims, kwargs...)
             return nameddimsarray_result(a, data, numerical_dims)
+        end
+    end
+end
+
+# 1 Arg, 2 Results
+for (mod, funs) in (
+    (:Base, (:findmax, :findmin)),
+)
+    for fun in funs
+        @eval function $mod.$fun(a::NamedDimsArray; dims=:, kwargs...)
+            numerical_dims = dim(a, dims)
+            data = $mod.$fun(parent(a); dims=numerical_dims, kwargs...)
+            map(data) do item  # value and index
+                nameddimsarray_result(a, item, numerical_dims)
+            end
         end
     end
 end
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -30,9 +30,7 @@ for (mod, funs) in (
 end
 
 # 1 Arg, 2 Results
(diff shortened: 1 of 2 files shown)


oxinabox:
According to CI this didn't workout.

fredcallaway:
Sorry, I've grown to distrust CI (often errors for reasons other than the new code) so I ignored this... Not sure how I managed to pass the tests locally—possibly a Revise issue. In any case, I have made extra sure that it works now.

However, I'm not sure we're ready to merge yet, for two reasons.

**1)** I noticed that indexing with a Matrix does not work as expected. For example, 

```julia
a = [10 20; 31 40]
@assert a[argmax(a; dims=1)] == maximum(a; dims=1)  # works
@assert nda[argmax(nda; dims=1)] == maximum(nda; dims=1)  # error :(

ERROR: ArgumentError: A 2 dimensional array, needs a 2-tuple of dimension names. Got: (:x,)
Stacktrace:
   [1] NamedDimsArray
     @ ~/.julia/dev/NamedDims/src/wrapper_array.jl:40 [inlined]
   [2] getindex(a::NamedDimsArray{(:x, :y), Int64, …}, raw_inds::NamedDimsArray{(:x, :y), CartesianIndex{2}, …})
     @ NamedDims ~/.julia/dev/NamedDims/src/wrapper_array.jl:162
   [3] top-level scope
     @ REPL[55]:1
```


**2)** I wonder if we really want to return a vanilla CartesianIndex for argmax and findmax because doing so throws away the dimension information. We could instead return a `NamedDimsIndex` (or NamedDimsCartesianIndex) that subtypes AbstractCartesianIndex. Not sure how easy it would be to adapt the existing indexing code to handle this type though.

mcabbott:
> indexing with a Matrix does not work as expected

This is a bug in indexing, I think, as it exists without the PR:
```julia
julia> argmax(a; dims=1)
1×2 Matrix{CartesianIndex{2}}:
 CartesianIndex(2, 1)  CartesianIndex(2, 2)

julia> nda[argmax(a; dims=1)]
ERROR: ArgumentError: A 2 dimensional array, needs a 2-tuple of dimension names. Got: (:x,)
```

>  could instead return a ... NamedDimsCartesianIndex

This sounds pretty complicated! Do you have a use case? Seems you would need to have two different arrays which you know have the same names but possibly not in the same order.

fredcallaway:
OK so the indexing bug is a separate issue.

My motivation for `NamedDimsCartesianIndex` is primarily that one might be interested in knowing what they keys are. For example, if a matrix is an objective function evaluated on a grid of two different parameters, I want to know the values of the parameters that maximize the objective, not the indices.

But ultimately, it's an easy one-liner to get that information back, so probably not worth investing all the effort. So perhaps we're good to go with this PR?

src/functions.jl
```diff
@@ -29,6 +29,21 @@ for (mod, funs) in (
     end
 end
 
+# 1 Arg, 2 Results
+for (mod, funs) in (
+    (:Base, (:findmax, :findmin)),
+)
```
oxinabox, line 35:
```suggestion
for (mod, funs) in ((:Base, (:findmax, :findmin)),)
```

oxinabox (approved):

PR #188, opened 2021-11-26
2-argument function to help const propagation in broadcasted_names()

### LINKED ISSUE #187: Type instability when broadcasting
I was trying to chase down allocations that appeared in my code when converting from `Array` to `NamedDimsArray`. I think I have found a case where we may be failing to get type-stable code. Here is what I think is a minimal example:
```
julia> using NamedDims, BenchmarkTools

julia> function example(a,b,c,d)
           @. a -= 0.5 * d * b * c
           return a
       end
example (generic function with 1 method)

julia> a = NamedDimsArray{(:z,)}(rand(5));

julia> b = NamedDimsArray{(:z,)}(rand(5));

julia> c = rand();

julia> d = rand();

julia> @btime example($a,$b,$c,$d)
  733.862 ns (10 allocations: 352 bytes)
5-element NamedDimsArray{(:z,), Float64, 1, Vector{Float64}}:
 -1085.0390848895406
 -3944.198097644197
 -4564.398164918389
 -3641.011715928582
 -5293.772958207846
```
Using `Cthulhu.@descend` suggests that the problem is return type not being inferred for `broadcasted_names(bs...)` from inside `copyto!()` here
https://github.com/invenia/NamedDims.jl/blob/2fa07ee7d94f34767db849942828f97ba0232b28/src/broadcasting.jl#L68-L80
even though when I then step into `broadcasted_names()`, the return type is `Core.Const((:z,))`!

I think the example is minimal, because if I remove even the `0.5 *` from `example()`, `@btime` reports zero allocations.

I was using julia-1.6.4 and NamedDims-0.2.41. Any suggestions for fixes or further debugging very welcome!

johnomotani:
Apparently less recursion helps the compiler to const-propagate.

Fixes #187.

### DIFF
--- a/src/broadcasting.jl
+++ b/src/broadcasting.jl
@@ -73,10 +73,18 @@ function Base.copyto!(dest::AbstractArray, bc::Broadcasted{NamedDimsStyle{S}}) w
 end
 
 broadcasted_names(bc::Broadcasted) = broadcasted_names(bc.args...)
-function broadcasted_names(a, bs...)
+function broadcasted_names(a, b)
     a_name = broadcasted_names(a)
-    b_name = broadcasted_names(bs...)
+    b_name = broadcasted_names(b)
     return unify_names_longest(a_name, b_name)
+end
+# Including two explicit arguments like this before starting recursion seems to help
+# const-propagation
+function broadcasted_names(a, b, cs...)
+    a_name = broadcasted_names(a)
+    b_name = broadcasted_names(b)
+    c_name = broadcasted_names(cs...)
+    return unify_names_longest(a_name, b_name, c_name)
 end
 broadcasted_names(a::AbstractArray) = dimnames(a)
 broadcasted_names(a) = tuple()
--- a/src/broadcasting.jl
+++ b/src/broadcasting.jl
@@ -78,7 +78,7 @@ function broadcasted_names(a, b)
     b_name = broadcasted_names(b)
     return unify_names_longest(a_name, b_name)
 end
-# Including two explicit arguments like this before starting recursion seems to help
+# Including two explicit arguments like this before starting recursion helps
 # const-propagation
 function broadcasted_names(a, b, cs...)
     a_name = broadcasted_names(a)
--- a/test/broadcasting.jl
+++ b/test/broadcasting.jl
@@ -149,13 +149,11 @@ function foo(a, b, c, d)
     @. a -= 0.5 * d * b * c
     return a
 end
-if VERSION >= v"1.4"
-    @testset "Regression test against #187 (allocations)" begin
-        # # https://github.com/invenia/NamedDims.jl/issues/187
-        a = NamedDimsArray{(:z,)}(rand(5))
-        b = NamedDimsArray{(:z,)}(rand(5))
-        c = rand()
-        d = rand()
-        @test 0 == @ballocated foo($a, $b, $c, $d)
-    end
+@testset "Regression test against allocations in broadcasting #187" begin
+    # https://github.com/invenia/NamedDims.jl/issues/187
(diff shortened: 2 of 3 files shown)


johnomotani:
I don't claim to understand why this works, if anyone does I'd be interested to know!

I originally made a different version that used meta-programming to generate an arbitrary number of `broadcasted_names()` functions with increasing numbers of arguments. For all the cases I tried, only the 2-argument case was needed to get const-propagation to work though. If someone has time to stress-test and find a case that breaks the 'simple' version in this commit, the meta-programming version was:
```
broadcasted_names(bc::Broadcasted) = broadcasted_names(bc.args...)
const unroll_size = 2 
# Create wrapper that handles calls with more arguments than unroll_size
let argnames = Tuple(Symbol("a$i") for i ∈ 1:unroll_size)
    eval(quote
         function broadcasted_names($(argnames...), bs...)
             a_name = broadcasted_names($(argnames...))
             b_name = broadcasted_names(bs...)
             return unify_names_longest(a_name, b_name)
         end
    end)
end
# Create unrolled functions that call broadcasted_names on each argument for numbers of
# arguments up to unroll_size
for n ∈ 2:unroll_size
    argnames = Tuple(Symbol("a$(i)") for i ∈ 1:n)
    names = Tuple(Symbol("a$(i)_name") for i ∈ 1:n)
    func_body = :() 
    for (argname, name) ∈ zip(argnames, names)
        func_body = quote
            $func_body
            $name = broadcasted_names($argname)
        end 
    end 
    eval(quote
        function broadcasted_names($(argnames...))
            $func_body
            return unify_names_longest($(names...))
        end 
    end)
end
broadcasted_names(a::AbstractArray) = dimnames(a)
broadcasted_names(a) = tuple()
```

mzgubic (approved):
LGTM, thanks for solving this! Could you bump the version please?

src/broadcasting.jl
```diff
@@ -73,10 +73,18 @@ function Base.copyto!(dest::AbstractArray, bc::Broadcasted{NamedDimsStyle{S}}) w
...
 broadcasted_names(bc::Broadcasted) = broadcasted_names(bc.args...)
-function broadcasted_names(a, bs...)
+function broadcasted_names(a, b)
     a_name = broadcasted_names(a)
-    b_name = broadcasted_names(bs...)
+    b_name = broadcasted_names(b)
     return unify_names_longest(a_name, b_name)
 end
+# Including two explicit arguments like this before starting recursion seems to help
+# const-propagation
```
oxinabox, line 82:
```suggestion
# Including two explicit arguments like this before starting recursion helps const-propagation
```

test/broadcasting.jl
```diff
@@ -144,3 +144,18 @@ end
...
+if VERSION >= v"1.4"
+    @testset "Regression test against #187 (allocations)" begin
+        # # https://github.com/invenia/NamedDims.jl/issues/187
+        a = NamedDimsArray{(:z,)}(rand(5))
+        b = NamedDimsArray{(:z,)}(rand(5))
+        c = rand()
+        d = rand()
+        @test 0 == @ballocated foo($a, $b, $c, $d)
+    end
+end
```
oxinabox, line 161:
```suggestion
@testset "Regression test against allocations in broadcasting #187" begin
    # https://github.com/invenia/NamedDims.jl/issues/187
    a = NamedDimsArray{(:z,)}(rand(5))
    b = NamedDimsArray{(:z,)}(rand(5))
    c = rand()
    d = rand()
    @test_modern 0 == @ballocated foo($a, $b, $c, $d)
end
```

oxinabox (approved):

oxinabox:
Thanks!

johnomotani:
Thanks for the speedy reviews!

PR #191, opened 2021-12-01
use ColumnNorm() and NoPivot() instead of Val(true) and Val(false) in qr calls

### LINKED ISSUE #190: MethodError: `qr` is ambiguous
Tests are currently broken because of 
```julia
  MethodError: qr(::NamedDimsArray{(:foo, :bar), Float64, 2, Matrix{Float64}}, ::Val{true}) is ambiguous. Candidates:
    qr(A::AbstractMatrix, ::Val{true}; kwargs...) in LinearAlgebra at deprecated.jl:70
    qr(nda::NamedDimsArray{L, T, N, A} where {N, A<:AbstractArray{T, N}}, args...; kwargs...) where {L, T} in NamedDims at /Users/mzgubic/JuliaEnvs/NamedDims.jl/src/functions_linearalgebra.jl:45
  Possible fix, define
    qr(::NamedDimsArray{L, T, 2, A} where A<:AbstractMatrix{T}, ::Val{true}) where {L, T}
```

~~I don't really know what changed to cause this ambiguity? I thought that LinearAlgebra versions are fixed to the julia version, while this seems to have happened across the versions (CI only shows mac on 1.0, but I suspect because it started the latest, can reproduce locally on 1.7).~~

oh, it's actually julia 1 jobs that failed, not julia 1.0

mzgubic:
closes #190

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.42"
+version = "0.2.43"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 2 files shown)


oxinabox:
This needs to be done in a way that does work for old versions
```
if !isdefined(@__MODULE__, :ColumnNorm)
    # Make work on old julia versions
    ColumnNorm() = true
end
```

mzgubic:
Thanks! I misread the CI as julia 1.0 failing (rather than 1) so was super confused as to why it started failing in the first place (see https://github.com/invenia/NamedDims.jl/issues/190) 😂

oxinabox (approved):

PR #196, opened 2022-03-12
fix depraction warning "Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead)"

KristofferC:


mzgubic (approved):

mzgubic:
Thanks!

PR #202, opened 2022-08-03
Add Cholesky factorisation

AlexRobson:
Presently `Cholesky` isn't supported with `NamedDims`. 

Adding this was made fairly easy with the `NamedFactorisation`. It throws a dimensionamismatch if the axes are not named the same. Also includes the suggested linting changes from JuliaFormatter for some lines that were not part of this MR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - "1.3"  # Would test on 1.0 (LTS), but that causes trouble at test time as test can't downgrade main deps in 1.0
           - "1"    # Latest Release
         os:
           - ubuntu-latest
@@ -30,11 +29,6 @@ jobs:
             arch: x86
           - os: windows-latest
             arch: x86
-        include:
-          # Add a 1.5 job because that's what Invenia actually uses
-          - os: ubuntu-latest
-            version: 1.5
-            arch: x64
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.49"
+version = "0.2.50"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.50"
+version = "0.3.0"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
@@ -20,7 +20,7 @@ ChainRulesTestUtils = "1"
 CovarianceEstimation = "0.2.4"
 Requires = "0.5, 1"
 Tracker = "0.2.2"
-julia = "1"
+julia = "1.6"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/src/functions_linearalgebra.jl
+++ b/src/functions_linearalgebra.jl
@@ -26,6 +26,7 @@ Base.propertynames(named::NamedFactorization; kwargs...) = propertynames(parent(
 Base.iterate(named::NamedFactorization{L, T, <:LU}) where {L, T} = (named.L, Val(:U))
 Base.iterate(named::NamedFactorization{L, T, <:LQ}) where {L, T} = (named.L, Val(:Q))
 Base.iterate(named::NamedFactorization{L, T, <:SVD}) where {L, T} = (named.U, Val(:S))
+Base.iterate(named::NamedFactorization{L, T, <:Cholesky}) where {L, T} = (named.L, Val(:U))
 function Base.iterate(
     named::NamedFactorization{L, T, <:Union{QR, LinearAlgebra.QRCompactWY, QRPivoted}}
 ) where {L, T}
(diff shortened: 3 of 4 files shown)


src/functions_linearalgebra.jl
```diff
@@ -82,6 +83,20 @@ function Base.getproperty(fact::NamedFactorization{L, T, <:LQ}, d::Symbol) where
     end
 end
 
+# cholesky
+function Base.getproperty(fact::NamedFactorization{L, T, <:Cholesky}, d::Symbol) where {L, T}
+    inner = getproperty(parent(fact), d)
+    n1, n2 = L
+    @assert n1 == n2 # Must be identical given square pdmat?
+    if d === :L
+        return NamedDimsArray{(n1, n2)}(inner)
```
AlexRobson, line 92: For cholesky, I assume that both the lower triangular and upper triangular being square matricies would correspond to the same axis.

src/functions_linearalgebra.jl
```diff
@@ -26,6 +26,7 @@ Base.propertynames(named::NamedFactorization; kwargs...) = propertynames(parent(
 Base.iterate(named::NamedFactorization{L, T, <:LU}) where {L, T} = (named.L, Val(:U))
 Base.iterate(named::NamedFactorization{L, T, <:LQ}) where {L, T} = (named.L, Val(:Q))
 Base.iterate(named::NamedFactorization{L, T, <:SVD}) where {L, T} = (named.U, Val(:S))
+Base.iterate(named::NamedFactorization{L, T, <:Cholesky}) where {L, T} = (named.L, Val(:U))
```
AlexRobson, line 29: IIUC this correctly, this should return:

```
r = rand(4,4); Σ = r' * r
_l, _u = cholesky(Σ)
cholesky(Σ).L == _l
cholesky(Σ).U == _u
```

Compare with lu::

```
_l, _u = lu(Σ)
lu(Σ).L == _l # True
lu(Σ).U == _u # True
```

test/functions_linearalgebra.jl
```diff
@@ -41,7 +45,7 @@ function baseline_tests(fact, identity)
             _named isa NamedDimsArray && @testset "Test name for dim $d" for d in 1:ndims(_named)
                 # Don't think it make sense for an factorization to produce properties with
                 # dimension sizes outside 1, 2 or 3
-                @test size(_named, d) in (1, 2, 3)
```
AlexRobson, line 44: I was confused with this test as it looks to be testing the size not the number of dims.

test/functions_linearalgebra.jl
```diff
@@ -10,13 +10,17 @@ if !isdefined(@__MODULE__, :ColumnNorm)
     NoPivot() = Val(false)
 end
 
-function baseline_tests(fact, identity)
+_test_data(::Val{:rectangle}) = [1.0 2 3; 4 5 6];
+_test_data(::Val{:pdmat}) =   [8.0  7.0  6.0  5.0; 7.0  8.0  6.0  6.0; 6.0  6.0  6.0  5.0; 5.0  6.0  5.0  5.0]
+_test_names(::Val{:rectangle}) = (:foo, :bar)
+_test_names(::Val{:pdmat}) = (:foo, :foo)
```
AlexRobson, line 16: Assumes that the axis of anything that is choleskied are the same?

src/functions_linearalgebra.jl
AlexRobson: This is working on julia 1 but failing on 1.{3/5} so I think i'm isunderstanding something here.

rofinn: I think that's fine. Though we should probably make this a minor release if you'd prefer to just pin the Julia version to 1.6 or higher.

src/functions_linearalgebra.jl
rofinn: Yes, that would be my assumption as well. That being said, there's nothing stopping someone from constructing a factorization type with a square matrix, but having different names. Should probably replace the `@assert` with a `DimensionMismatch` error though.

rofinn: Could probably also rewrite this block as:

```
return d in (:L, :U) ? NamedDimsArray{(n1, n2)}(inner) : inner
```

rofinn (approved):
Seems mostly fine apart from cleaning up the `@assert` and bumping the release for Julia >=1.6.

src/functions_linearalgebra.jl
AlexRobson: I realised actually we don't want the dimcheck at the point of getproperty, but at the point of constructing the `NamedFactoriztion`. I've added a new constructor for this

test/functions_linearalgebra.jl
```diff
@@ -143,6 +152,21 @@ end
...
+
+    # Explicit `dimnames` tests for readability
+    nda = NamedDimsArray{(:foo, :foo)}(_test_data(Val{:pdmat}()))
+    nda_mismatch = NamedDimsArray{(:foo, :bar)}(_test_data(Val{:pdmat}()))
+    x = cholesky(nda)
+    @test size(x) == size(parent(x))
+    @test dimnames(x.L) == (:foo, :foo)
+    @test dimnames(x.U) == (:foo, :foo)
+
+    @test_throws DimensionMismatch cholesky(nda_mismatch)
```
AlexRobson, line 167: now throw an error if there is a dim-mismatch.

.github/workflows/CI.yml
```diff
@@ -30,11 +29,6 @@ jobs:
             arch: x86
           - os: windows-latest
             arch: x86
-        include:
```
AlexRobson, line 33: drop julia < 1.6 in CI

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.49"
+version = "0.3.0"
```
AlexRobson, line 4: major bump due to constrainting julia to 1.6 and above

src/functions_linearalgebra.jl
```diff
@@ -82,8 +85,24 @@ function Base.getproperty(fact::NamedFactorization{L, T, <:LQ}, d::Symbol) where
...
 end
 
+# cholesky
+
+function Base.getproperty(fact::NamedFactorization{L,T,<:Cholesky}, d::Symbol) where {L,T}
+    inner = getproperty(parent(fact), d)
+    n1, n2 = L
+    return d in (:L, :U) ? NamedDimsArray{(n1, n2)}(inner) : inner
+end
+function NamedFactorization{L}(fact::Cholesky{T}) where {L,T}
```
AlexRobson, line 95: Define a new constructor to handle the dimension mismatch

rofinn (approved):
Still looks fine to me

Project.toml
ericphanson: in the future, this should be a patch bump to be non-breaking (since the package is pre-1.0) unless one intends to support backports: https://github.com/SciML/ColPrac/pull/20/files

PR #203, opened 2022-08-04
Support `isposdef`

rofinn:


test/functions_linearalgebra.jl
```diff
@@ -199,3 +199,15 @@ end
         NamedDimsArray{(:A, :B)}(r1), NamedDimsArray{(:NotA,)}(b)
     )
 end
+
+@testset "isposdef" begin
+    A = [1 2; 2 50]
+
+    @test isposdef(A)
```
oxinabox, line 206:
```suggestion
    @assert isposdef(A)
```
When it can't be fixed by changing something `src/`, I think it is better to use `@assert`.
It is just an assumption of the tests

test/functions_linearalgebra.jl
```diff
@@ -199,3 +199,15 @@ end
...
+    @test isposdef(NamedDimsArray{(:foo, :foo)}(A))
+
+    B = Float64.(A)
+    @test isposdef!(B)
```
oxinabox, line 210:
```suggestion
    @assert isposdef!(B)
```

oxinabox (approved):
minor quibble

test/functions_linearalgebra.jl
rofinn: Yeah, I guess that makes sense.

PR #204, opened 2022-08-05
Release 1.0 ?

oxinabox:
This is pretty much stable.
This release will be nonbreaking and the same as the last 0.3.x release which is frustrating but still better I think than leaving this 0.x forever.

Though maybe this isn't as stable as i thought, why did we release 0.3?

Invenia's new  internal policy is to create packages at 1.0.0 and never use  0.x.y versioning

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NamedDims"
 uuid = "356022a1-0364-5f58-8944-0da4b18d706f"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.3.1"
+version = "1.0.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


mzgubic:
https://github.com/invenia/NamedDims.jl/pull/202/files#r937572855

we dropped support for Julia 1.5 and below. Is that in line with SemVer?

oxinabox:
Dropping julia compatibility basically never can be considered breaking by SemVer.
Because Pkg will prevent anything not compatible with this version of julia from ever installing.

Convention for packages that are post v1.0.0 is to use a minor release so that  patch releases can squeeze in any backports easily. ( and that is enforced at the registry CI IIRC)
Pre v1.0.0 it is just another nonbreaking release.

mzgubic (approved):

PR #211, opened 2022-10-30
infer reduce(*cat)

aplavin:
`_named_*cat` funcs rely on constant propagation of names, so they need inlining to infer


mzgubic (approved):
Thanks!

